PR #10, opened 2022-12-01
Send/Receive messages

nakajima:
Extracted from https://github.com/xmtp/xmtp-ios/pull/6.
Relies on https://github.com/xmtp/xmtp-ios/pull/9.


Sources/XMTP/Client.swift
```diff
@@ -38,7 +38,27 @@ class Client {
...
+
+			var apiClient = apiClient
+			apiClient.setAuthToken(authToken)
+
+			do {
+				try await apiClient.publish(envelopes: [
+					Envelope(topic: .userPrivateStoreKeyBundle(account.address), timestamp: Date(), message: try encryptedKeys.serializedData()),
+				])
+			} catch {
+				// Unable to save keys
```
elisealix22, line 58: Maybe worth bubbling this up or returning nil if it breaks things? Or can we move on without it?

Sources/XMTP/Codecs/TextCodec.swift
```diff
@@ -0,0 +1,43 @@
...
+//  TextCodec.swift
+//
+//
+//  Created by Pat Nakajima on 11/28/22.
+//
+
+import Foundation
+import XMTPProto
+
+let ContentTypeText = ContentTypeID(authorityID: "xmtp.org", typeID: "text", versionMajor: 1, versionMinor: 0)
```
elisealix22, line 11: 🔥

elisealix22 (approved):
Wait, what the heck? This is the whole SDK  😆 . Just one question / suggestion on the saving of keys in creating the client. Otherwise, this is epic.

Sources/XMTP/Client.swift
nakajima: Ya, I agree. I've removed this try/catch.

PR #11, opened 2022-12-05
Integrate SDK into sample App

nakajima:
This PR starts exposing the public interface for the SDK. In the course of doing so, I realized that V2 messages weren't getting encoded/decoded properly, so I fixed that and added tests.

(Sidenote: I've been doing a lot of testing with a local node and a script in xmtp-js so I can verify the stuff I'm generating works properly with that. At some point it might be nice to try to standardize a format for end to end integration testing like that.)

### DIFF
--- a/Sources/XMTP/Conversation.swift
+++ b/Sources/XMTP/Conversation.swift
@@ -5,6 +5,7 @@
 //  Created by Pat Nakajima on 11/28/22.
 //
 
+import Foundation
 import XMTPProto
 
 public enum Conversation {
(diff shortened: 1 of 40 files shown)


Sources/XMTP/Conversation.swift
```diff
@@ -5,8 +5,46 @@
 //  Created by Pat Nakajima on 11/28/22.
 //
 
+import Foundation
 import XMTPProto
 
-enum Conversation {
+public enum Conversation {
+	// TODO: It'd be nice to not have to expose these types as public, maybe we make this a struct with an enum prop instead of just an enum
```
elisealix22, line 12: Sounds good! But wouldn't stress this too much either imo.

elisealix22 (approved):
LGTM! Very cool that the example is in the same repo. I think that will make it easier for everyone and matches what I'm used to in OSS.

PR #12, opened 2022-12-06
Enhance Conversation APIs

nakajima:


Sources/XMTP/Messages/Invitation.swift
```diff
@@ -8,17 +8,15 @@
...
 extension InvitationV1 {
-	@available(iOS 16.0, *)
 	static func createRandom(context: InvitationV1.Context? = nil) throws -> InvitationV1 {
 		let context = context ?? InvitationV1.Context()
 		let randomBytes = try Crypto.secureRandomBytes(count: 32)
-		let regex = #/=*$/#
 		let randomString = Data(randomBytes).base64EncodedString()
-			.replacing(regex, with: "")
-			.replacingOccurrences(of: "/", with: "-")
+			.replacingOccurrences(of: "=*$", with: "", options: .regularExpression)
```
nakajima, line 18: Changed this so we can stop relying on iOS 16 only regex stuff.

neekolas (approved):
This is great stuff @nakajima. Not qualified to evaluate the Swiftiness, but the business logic checks out.

Sources/XMTP/Conversations.swift
```diff
@@ -11,6 +11,61 @@ import XMTPProto
...
+		}
+
+		// See if we have a v2 conversation
+		for sealedInvitation in try await listInvitations() {
+			if !sealedInvitation.involves(contact) {
+				continue
+			}
+
+			let invite = try sealedInvitation.v1.getInvitation(viewer: client.keys)
+			if invite.context.conversationID == context?.conversationID, invite.context.conversationID != "" {
```
nakajima, line 39: I'm a bit curious where `conversationID` gets used? I see the JS SDK checks for it in a bunch of places but I didn't see any places where it gets a real value? Is it something clients are meant to specify?

neekolas: `newConversation`

Sources/XMTP/ApiClient.swift
```diff
@@ -9,8 +9,18 @@ import GRPC
 import XMTPProto
 
 typealias PublishResponse = Xmtp_MessageApi_V1_PublishResponse
+typealias QueryResponse = Xmtp_MessageApi_V1_QueryResponse
+
+protocol ApiClient {
+	var environment: Environment { get }
+	init(environment: Environment, secure: Bool) throws
+	func setAuthToken(_ token: String)
```
elisealix22, line 17: Today I learned about omitting argument labels in Swift tyty! Makes total sense here too.

Sources/XMTP/Conversations.swift
elisealix22: Yeah, [here's an example](https://github.com/lensterxyz/lenster/blob/30d458b792726ff33430ae1012b8385b9f4f656b/apps/web/src/components/utils/hooks/useGetConversation.tsx#L36) of it being used in the wild. It allows consumers of the SDK to do things like filter out all other convos except the ones that match the convo id. In XMTP's example web app, we are acting like a "universal inbox" so we don't do any client-side filtering by convo id there.

Sources/XMTP/Messages/Invitation.swift
elisealix22: Amazing.

Sources/XMTP/Messages/MessageV1.swift
```diff
@@ -82,7 +82,7 @@ extension MessageV1 {
 
 	var sentAt: Date {
 		// swiftlint:disable force_try
-		try! Date(timeIntervalSince1970: Double(header.timestamp) / 1_000_000)
+		try! Date(timeIntervalSince1970: Double(header.timestamp / 1000))
```
elisealix22, line 85: Just to double check: Is this correct here? I would assume the header timestamp is is in nanoseconds here and that the timeIntervalSince1970 would be milliseconds.

Tests/XMTPTests/ConversationTests.swift
```diff
@@ -0,0 +1,109 @@
...
+
+		let newConversations = try await aliceClient.conversations.list()
+		XCTAssertEqual(1, newConversations.count, "already had conversations somehow")
+	}
+
+	func testCanFindExistingV1Conversation() async throws {
+		let encoder = TextCodec()
+		let encodedContent = try encoder.encode(content: "hi alice")
+
+		let date = Date().advanced(by: -1_000_000)
```
elisealix22, line 55: Similar time q: What's the purpose of it being negative here? I'm reading this as if we're going back in time by 1_000_000 seconds based on [advancedBy / TimeInterval docs](https://developer.apple.com/documentation/foundation/date/3329237-advanced) but idk if that's correct.

elisealix22 (approved):
A couple of questions around the time conversion (mostly for my understanding), but other than that LGTM!

Sources/XMTP/Messages/MessageV1.swift
elisealix22: Related (and feel free to ignore!) but we ended up adding nanosecond / millisecond conversion helpers in both `xmtp-js` and [Flutter](https://github.com/xmtp/xmtp-flutter/pull/10/files#diff-ad10af43adb21e81d72e92a82c6bd81c04ed9dca3c18cacbd585cd38d50c0e7fR6) in case this inspires you or is helpful at all 😆

nakajima: 🤦 My test was wrong here, once I used the millisecond helper I had already added ages ago, I was able to remove this. Great catch!

Tests/XMTPTests/ConversationTests.swift
nakajima: I just wanted a specific date in the past to test with to make sure it was carried through in the tests, rather than using the current `Date()`, which could lead to things "Just working" even if I was newing up a fresh one somewhere I shouldn't be. I could see how using `1_000_000` would be confusing though so I'll change the magic number and leave a comment to explain what's going on here.

Sources/XMTP/Conversations.swift
nakajima: Gotcha, thanks!

Tests/XMTPTests/ConversationTests.swift
mkobetic: Or maybe just name the variable `someTimeAgo` or something.

nakajima: > Or maybe just name the variable someTimeAgo or something.

I like that.

PR #14, opened 2022-12-07
Integrate message streaming and newConversation into example app

nakajima:
https://user-images.githubusercontent.com/483/206316116-27c9cfd3-a840-44b6-acdb-e48ab063b179.mov


Sources/XMTP/ApiClient.swift
```diff
@@ -13,8 +13,8 @@ typealias QueryResponse = Xmtp_MessageApi_V1_QueryResponse
 typealias SubscribeRequest = Xmtp_MessageApi_V1_SubscribeRequest
 
 protocol ApiClient {
-	var environment: Environment { get }
-	init(environment: Environment, secure: Bool) throws
+	var environment: XMTPEnvironment { get }
```
nakajima, line 16: Renamed this because `Environment` was clashing with SwiftUI's `@Environment`.

Sources/XMTP/Client.swift
```diff
@@ -47,7 +51,7 @@ public class Client {
 		let privateKeyBundleV1 = try await loadOrCreateKeys(for: account, apiClient: apiClient)
 
 		let client = try Client(address: account.address, privateKeyBundleV1: privateKeyBundleV1, apiClient: apiClient)
-		try await client.publishUserContact()
+		try await client.ensureUserContactPublished()
```
nakajima, line 54: Brought over the logic to not constantly be publishing new contact bundles here.

XMTPiOSExample/XMTPiOSExample/Views/LoggedInView.swift
```diff
@@ -8,15 +8,21 @@
...
+}
+
 struct LoggedInView: View {
 	var client: XMTP.Client
 
+	@StateObject var environmentCoordinator = EnvironmentCoordinator()
+
 	var body: some View {
-		NavigationView {
+		NavigationStack(path: $environmentCoordinator.path) {
```
nakajima, line 21: Using `NavigationStack` instead of `NavigationPath` here means the demo app is iOS 16 only. Feels worth it to me since it's just for educational purposes and using the old navigation stuff ends up with piles of hacks on piles of hacks.

XMTPiOSExample/XMTPiOSExample/Views/MessageComposerView.swift
```diff
@@ -30,12 +35,17 @@ struct MessageComposerView: View {
 	}
 
 	func send() {
+		if text.trimmingCharacters(in: .whitespacesAndNewlines) == "" {
```
nakajima, line 38: Is there any sort of "server side" validation to ensure we don't have empty messages or is that not really a concern?

elisealix22: There is not currently and not sure if it's intentional or not. I've been [adding checks in the frontend](https://github.com/lensterxyz/lenster/blob/1d358d08ce150d20a281f2cd728d8d54754c4f06/apps/web/src/components/Messages/Composer.tsx#L24) to disable the send button until there's content (trimming whitespace like you're doing).

@mkobetic not urgent, but do you know if it's intentional that we allow blank messages?

elisealix22 (approved):
Love it!

PR #16, opened 2022-12-12
E2E Testing

nakajima:
Start introducing E2E testing abilities. I also used those tests to fix "key not signed by wallet" errors when communicating between the Swift and the JS sdk. Right now I'm just being super explicit everywhere about which field the signatures are in, if v1 is deprecated I think this could get loosened a bit.

I was going to enable these in CI but ran into a bunch of networking issues I still need to figure out (it could be the order of service launching in docker compose but I'm not quite sure). Either way this PR has been open long enough so I wanted to land it then look at cleanups in follow ups.

### DIFF
--- a/Sources/XMTP/Messages/Signature.swift
+++ b/Sources/XMTP/Messages/Signature.swift
@@ -73,19 +73,6 @@ extension Signature {
 		}
 	}
 
-	mutating func ensureEcdsaSignature() {
-		switch union {
-		case let .walletEcdsaCompact(walletEcdsa):
-			var ecdsa = Signature.ECDSACompact()
-			ecdsa.bytes = walletEcdsa.bytes
-			ecdsa.recovery = walletEcdsa.recovery
-			ecdsaCompact = ecdsa
-			union = .ecdsaCompact(ecdsa)
-		case .ecdsaCompact(_), .none:
-			return
-		}
-	}
-
 	mutating func ensureWalletSignature() {
 		switch union {
 		case let .ecdsaCompact(ecdsa):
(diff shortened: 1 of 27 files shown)


Sources/XMTP/Messages/Signature.swift
```diff
@@ -63,11 +63,40 @@ extension Signature {
...
+			return ecdsa.bytes + [UInt8(Int(ecdsa.recovery))]
+		case let .walletEcdsaCompact(ecdsa):
+			return ecdsa.bytes + [UInt8(Int(ecdsa.recovery))]
+		case .none:
+			return Data()
 		}
+	}
 
-		return ecdsaCompact.bytes + [UInt8(Int(ecdsaCompact.recovery))]
+	mutating func ensureEcdsaSignature() {
```
elisealix22, line 76: Is this still necessary after this change upstream? https://github.com/xmtp/xmtp-node-go/pull/193

Asking because I do see a lot of lets turning into vars which is totally fine if it's necessary but seams like it would be cleaner if it was fixed in the go repo!

elisealix22 (commented):
LOVE the e2e tests ❤️ . Just one q on if the modifying of sigs is still necessary with a recent change.

Sources/XMTP/Messages/Signature.swift
nakajima: Good question, I didn't see that PR had landed so I'll pull it down and see how stuff works in the new world.

nakajima: @elisealix22 I was about to remove a bunch of those! Thanks for giving me the heads up about that PR landing.

elisealix22 (approved):
😍

PR #19, opened 2022-12-15
Update README

nakajima:
Here's a first draft at adapting the xmtp-js for this package.

[Rendered Markdown Here](https://github.com/xmtp/xmtp-ios/blob/c4ba09bd13818ef0d4623a5607d628d32c0bb426/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ let newConversation = try await client.conversations.newConversation(with: "0x3F
 
 ### Send messages
 
-To be able to send a message, the recipient must have already started a client at least once and consequently advertised their key bundle on the network. Messages are addressed using account addresses. The message payload must be a plain string.
+To be able to send a message, the recipient must have already created a client at least once and consequently advertised their key bundle on the network. Messages are addressed using account addresses. The message payload must be a plain string.
 
 **Note: Other types of content are currently not supported.**
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,278 @@
-# xmtp-ios
-XMTP client SDK for iOS applications written in Swift.
+# XMTP-iOS
 
-COMING SOON! This is a **work in progress** and should not be used yet. Follow along in the [tracking issue](https://github.com/xmtp/xmtp-ios/issues/7) for updates on progress.
+This SDK is in Developer Preview status. Software in this status:
+
+- Is not formally supported
+- Will change without warning
+- May not be backward compatible
+- We do NOT recommend using Developer Preview software in production apps.
+
+Specifically, this SDK is missing this functionality:
+
+- Specifying `apiUrl`, `keyStoreType`, `codecs`, `maxContentSize` and `appVersion` when creating a `Client`
+- Content types other than text
+- Streaming all messages from all conversations
+- Message content compression
+
+Follow along in the [tracking issue](https://github.com/xmtp/xmtp-ios/issues/7) for updates.
+
+![Test](https://github.com/xmtp/xmtp-ios/actions/workflows/test.yml/badge.svg)
+![Lint](https://github.com/xmtp/xmtp-ios/actions/workflows/lint.yml/badge.svg)
+
+![x-red-sm](https://user-images.githubusercontent.com/510695/163488403-1fb37e86-c673-4b48-954e-8460ae4d4b05.png)
+
+**XMTP client SDK for iOS applications**
+
+`xmtp-ios` provides a Swift implementation of an XMTP client for use with iOS applications.
+
+Build with `xmtp-ios` to provide messaging between blockchain wallet addresses, delivering on use cases such as wallet-to-wallet messaging and dapp-to-wallet notifications.
+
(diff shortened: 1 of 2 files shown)


nplasterer (approved):
🔥

README.md
```diff
@@ -127,11 +124,11 @@ for conversation in allConversations {
...
 ```
 
-#### Listen for new conversations
+### Listen for new conversations
 
-You can also listen for new conversations being started in real-time. This will allow applications to display incoming messages from new contacts.
+You can also listen for new conversations being started in real-time. This will allow apps to display incoming messages from new contacts.
 
-_Warning: this stream will continue infinitely. To end the stream, break from the loop`_
+**Warning: This stream will continue infinitely. To end the stream, break from the loop`**
```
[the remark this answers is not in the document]
nakajima: Should definitely be a period, this was a bad copy/paste.

```suggestion
**Warning: This stream will continue infinitely. To end the stream, break from the loop.**
```

README.md
```diff
@@ -1,60 +1,57 @@
...
 
-![Test](https://github.com/xmtp/xmtp-ios/actions/workflows/test.yml/badge.svg)
-![Lint](https://github.com/xmtp/xmtp-ios/actions/workflows/lint.yml/badge.svg)
-![Status](https://camo.githubusercontent.com/5bb5892781bbf711c7fe5eba3328e9e15a767de87c587d3fb65f2fd7e1f4ae72/68747470733a2f2f696d672e736869656c64732e696f2f62616467652f50726f6a6563745f5374617475732d446576656c6f7065725f507265766965772d726564)
+To learn more about XMTP and get answers to frequently asked questions, see [FAQ about XMTP](https://xmtp.org/docs/dev-concepts/faq).
 
 ![x-red-sm](https://user-images.githubusercontent.com/510695/163488403-1fb37e86-c673-4b48-954e-8460ae4d4b05.png)
 
-**XMTP client SDK for iOS applications**
+## Example app
```
jhaaaa, line 31: I thought this heading text might capture dev attention - but is it an accurate description of what's in the example project? =) If not, we should change it to "Example project" in the heading and the link text below.

nakajima: I guess what do we consider an "app" vs a "project?" You can install and run the example so by that metric it's an app, if not a very fully featured one.

jhaaaa: Thank you, Pat! I think we should go with "app" then. =) Perhaps we can set some expectations about the robustness of the app by making the edit below?

README.md
```diff
@@ -1,4 +1,282 @@
...
+
+Follow along in the [tracking issue](https://github.com/xmtp/xmtp-ios/issues/7) for updates.
+
+To learn more about XMTP and get answers to frequently asked questions, see [FAQ about XMTP](https://xmtp.org/docs/dev-concepts/faq).
+
+![x-red-sm](https://user-images.githubusercontent.com/510695/163488403-1fb37e86-c673-4b48-954e-8460ae4d4b05.png)
+
+## Example app
+
+For a demonstration of the core concepts and capabilities of the `xmtp-ios` client SDK, see the [Example app project](https://github.com/xmtp/xmtp-ios/tree/main/XMTPiOSExample/XMTPiOSExample).
```
jhaaaa, line 33:
```suggestion
For a basic demonstration of the core concepts and capabilities of the `xmtp-ios` client SDK, see the [Example app project](https://github.com/xmtp/xmtp-ios/tree/main/XMTPiOSExample/XMTPiOSExample).
```

README.md
nakajima: Yep, works for me!

README.md
```diff
@@ -145,17 +142,17 @@ for try await conversation in client.conversations.stream() {
...
 
 ```swift
 let newConversation = try await client.conversations.newConversation(with: "0x3F11b27F323b62B159D2642964fa27C46C841897")
 ```
 
-#### Sending messages
+### Send messages
 
-To be able to send a message, the recipient must have already started their Client at least once and consequently advertised their key bundle on the network. Messages are addressed using wallet addresses. The message payload must be a plain string.
+To be able to send a message, the recipient must have already started a client at least once and consequently advertised their key bundle on the network. Messages are addressed using account addresses. The message payload must be a plain string.
```
nakajima, line 155: I wonder if `created` would be a better term to use here than `started`, since the method in the code is `Client.create` and the word `start` doesn't really appear anywhere in the SDK.

```suggestion
To be able to send a message, the recipient must have already created a client at least once and consequently advertised their key bundle on the network. Messages are addressed using account addresses. The message payload must be a plain string.
```

jhaaaa: Yes, please - this is a great callout, @nakajima! Let's do it - and I'll note it as an edit to make for other README work.

jhaaaa (approved):
Thank you for this AMAZING README and for so graciously considering my edits, @nakajima - looks great to me!

nakajima:
Thanks for all your help @jhaaaa!

PR #22, opened 2022-12-20
Improve docs

nakajima:
Adds typedocs to more things. This PR also makes a couple types private that didn't need to be public. These types weren't used by anything in the example app and shouldn't affect any users of the SDK. They were just cluttering the generated docs so I made them private.

Also I think these could all be fleshed out more, but at least now the generated docs won't be blank. I can also update this PR to try to get the package docs listed on https://blog.swiftpackageindex.com/posts/auto-generating-auto-hosting-and-auto-updating-docc-documentation/, unless we want to host them elsewhere in which case we can figure that out in a follow up PR.


elisealix22 (approved):
💚  Reads great to me!

Issue #29, opened 2023-01-04
Feature Request: WalletConnect should not be part of the library


Cordt:
**Context:**
As most apps (including lentil) will likely use WalletConnect to allow users to sign messages with their wallet, they will implement specific flows to interact with WalletConnect that fits their app's UX well and supports a certain set of wallets.

**Problems:**
When integrating the XMTP SDK, I am facing two issues
1. I cannot use a different version of the WalletConnect SDK than the one used by xmtp-ios, which in turn means I am also bound to the specific set of features and capabilities of that particular version
  - In order to make the build work, I had to remove the WalletConnect package from lentil so that the one from xmtp-ios is used instead
3. The SDK is tightly coupled with WalletConnect, which both makes it hard to mock the interfaces and breaks the flow with WalletConnect

**Suggested solution:**
- Signing should be independent of the mechanism used to do so (and I believe this is the direction you want to go anyway, looking at the `SigningKey` protocol)
  - Using a different implementation than the one provided by the Library for WalletConnect was quite challenging, as there are hidden requirements on the implementation like
    - how the sign message is implemented specifically: https://github.com/Cordt/Lentil/blob/5858713b4078c003b8c5f1c91995280c82a9c6e7/Lentil/Wallet/WalletConnector%2BSigningKey.swift#L14
    - and that the connection only works wrapping the addresses used in Ethereum Addresses: https://github.com/Cordt/Lentil/blob/5858713b4078c003b8c5f1c91995280c82a9c6e7/Lentil/Wallet/WalletConnector.swift#L149
- Signing requests should be communicated as Async Streams, to allow opening a wallet as often as necessary
  - Right now, XMTP requires two signatures when setting up a given address for XMTP initially
  - An Async Stream would allow to open the wallet twice, as WC currently doesn't seem to allow queueing multiple signatures
  - I already built an Async Stream as an abstraction on top of WalletConnect, so that I can fully control the flow of opening the wallet, connecting sessions and handling signatures - happy to help build something similar in the XMTP iOS SDK

nakajima:
> Signing should be independent of the mechanism used to do so (and I believe this is the direction you want to go anyway, looking at the SigningKey protocol)

Ya, I think you're right. That's the intention of the SigningKey protocol. We can remove WalletConnect from the SDK itself and then just show an example of how to integrate it with the example app. 

> Signing requests should be communicated as Async Streams, to allow opening a wallet as often as necessary

Cool, I dig that idea.

[...]

PR #32, opened 2023-01-06
Remove wallet connect as an SDK dependency

nakajima:
See https://github.com/xmtp/xmtp-ios/issues/29.


elisealix22 (approved):
Love this! Makes sense for it to live in the example.

Issue #30, opened 2023-01-04
Ability to export / load XMTP keys to store on secure enclave


nmalzieu:
In the JS Xmtp Client, there is a way to export the XMTP private key to store in a secure enclave for instance (https://github.com/xmtp/xmtp-js#manually-handling-private-key-storage)
Is there a way to do the same with the Swift XMTP Client currently? This part is missing from the doc

PR #33, opened 2023-01-06
Allow Client to be created from a V2 public key bundle

nakajima:
This PR introduces the functionality from the JS SDK described here: https://github.com/xmtp/xmtp-js#manually-handling-private-key-storage.

Should fix https://github.com/xmtp/xmtp-ios/issues/30.

TODO

- [x] ~~The JS SDK exposes a `PrivateKeyBundleV1`, but it seems like we should move away from that so this PR adds behavior that deals with V2 bundles. Is that the right call?~~ Decided to go with v1.
- [x] Documentation update

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,24 @@ import XMTP
 let client = try await Client.create(account: account)
 ```
 
+### Creating a client from saved keys
+
+You can save your keys from the client via the `v1keys` property:
+
+```swift
+// Create the client with a `SigningKey` from your app
+let client = try await Client.create(account: account)
+
+// Save the keys somewhere safe
+let keys = client.v1keys
+```
+
+Once you have those keys, you can create a new client with `Client.from`:
+
+```swift
+let client = try Client.from(bundle: keys)
+```
+
 ### Configure the client
 
 You can configure the client's network connection and key storage method with these optional parameters of `Client.create`:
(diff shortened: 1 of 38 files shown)


README.md
```diff
@@ -79,6 +79,24 @@ import XMTP
 let client = try await Client.create(account: account)
 ```
 
+### Creating a client from saved keys
+
+You can save your keys from the client via the `v1keys` property:
+
+```swift
+// Create the client with a `SigningKey` from your app
+let client = try await Client.create(account: account)
```
elisealix22, line 88: Suggestion: It might be worth adding the environment to the example because we've run into many issues in the past where people forget dev keys != prod keys.

README.md
```diff
@@ -79,6 +79,24 @@ import XMTP
...
+
+// Save the keys somewhere safe
+let keys = client.v1keys
+```
+
+Once you have those keys, you can create a new client with `Client.from`:
+
+```swift
+let client = try Client.from(bundle: keys)
```
elisealix22, line 97: And then also including the environment here!

elisealix22 (approved):
One small suggestion in the README, but this is epic! Excited to use it!

README.md
```diff
@@ -79,6 +79,24 @@ import XMTP
...
+### Creating a client from saved keys
+
+You can save your keys from the client via the `v1keys` property:
+
+```swift
+// Create the client with a `SigningKey` from your app
+let client = try await Client.create(account: account)
+
+// Save the keys somewhere safe
+let keys = client.v1keys
```
neekolas, line 91: Would be helpful to show to how to serialize these keys as well, since the most common use-case is going to be to store them somewhere which doesn't accept a class.

PR #36, opened 2023-01-17
Remove CryptoSwift dependency and resolve warnings

elisealix22:
I noticed an unused dependency warning for `CryptoSwift`. This PR removes the `CryptoSwift` dependency and cleans up the warnings in the library package! Let me know if the warnings should be ignored instead of resolved anywhere of course.


nakajima (approved):
I think I was trying to use some stuff out of CryptoSwift before but then moved away from it. Thanks for cleaning this (and the other stuff) up!

Issue #39, opened 2023-01-18
WalletConnect: No pre-key found


elisealix22:
In using the example app to connect with WC, I noticed all my conversations on the dev network error with `Error loading invitations: noPreKeyFound`. These conversations were created on https://xmtp.vercel.app/ so makes me think there's a mismatch in the Account's keys.

<img width="589" alt="Screenshot 2023-01-18 at 1 21 33 PM" src="https://user-images.githubusercontent.com/556051/213263028-a6b8e865-e6bd-4984-9a61-e2c5115f4be5.png">

PR #40, opened 2023-01-18
Fix loading saved keys

nakajima:
This PR fixes two issues with loading private key bundles:

- The topic for the private store needed to end with `/key_bundle`
- The recovery byte for the "Enable Identity" signature generated by the JS sdk was [either 27 or 28](https://github.com/ethers-io/ethers.js/blob/bf0b468490cb293cd916e4fff06e0909273719e6/packages/bytes/src.ts/index.ts#L480), but the secp256k1 library we're using expected 0 or 1. This was usually fine since those two values basically mean the same thing so recovery was working fine. The problem was when we were using the signature data as a secret for encrypting/decrypting.

Fixes https://github.com/xmtp/xmtp-ios/issues/39.

### DIFF
--- a/Sources/XMTP/Client.swift
+++ b/Sources/XMTP/Client.swift
@@ -127,9 +127,6 @@ public class Client {
 		let topics: [Topic] = [.userPrivateStoreKeyBundle(account.address)]
 		let res = try await apiClient.query(topics: topics, pagination: nil)
 
-		print("Loading private keys for \(account.address)")
-		print("Got envelopes \(res.envelopes)")
-
 		for envelope in res.envelopes {
 			let encryptedBundle = try EncryptedPrivateKeyBundle(serializedData: envelope.message)
 			let bundle = try await encryptedBundle.decrypted(with: account)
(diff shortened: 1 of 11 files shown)


Sources/XMTP/Client.swift
```diff
@@ -119,16 +127,13 @@ public class Client {
 		let topics: [Topic] = [.userPrivateStoreKeyBundle(account.address)]
 		let res = try await apiClient.query(topics: topics, pagination: nil)
 
+		print("Loading private keys for \(account.address)")
```
elisealix22, line 130: Were these meant to be debug only? Just checking because of the envelopes one seems like it might be noisy but I don't feel strongly

elisealix22 (approved):
Wow, thank you for the quick fix!

Sources/XMTP/Client.swift
nakajima: Whoops, didn't mean to leave this in.

neekolas:
Oof. Yeah, both of those issues have bit me in the past. We should do a better job of documenting those footguns.

Issue #31, opened 2023-01-05
Add a way to decode a single message


nmalzieu:
In the Javascript SDK, if we have a `conversation` instance and an encoded message, we can decode it using `conversation.decodeMessage`

Can we do the same with the SWIFT SDK? It would be useful to decode incoming notifications in an iOS Notification Extension

Bonus: if we can export / import keys (like https://github.com/xmtp/xmtp-ios/issues/30) it would also be amazing to be able to export / import saved conversations! And when importing conversations, being able to decode a notification without any access to the network, because sometimes if the connection is not good we still receive a notification, and the notification extension only has so much time to decode it

nakajima:
There is not yet a way to do this, I think it makes sense to add though. I can look at introducing it after https://github.com/xmtp/xmtp-ios/pull/26 lands.

nmalzieu:
Thanks @nakajima ! Can't wait to have this for Converse. I think it is the only way to have "real notifications" in a native XMTP app

iOS Notification Extension are made exactly for this: decode encoded content from notifications. Background notifications that are supposed to wake up the app are not a good solution because it doesn't work if the user killed the app. Only iOS Notification Extensions work 100% of the time!

But they don't have a lot of time to do their work before being killed by the system
Since we have the encoded message in the notification, being able to decode a single message makes sense!

nakajima:
Now that https://github.com/xmtp/xmtp-ios/pull/41 has merged, can we close sthis?

nmalzieu:
@nakajima this is not an issue anymore to me, you can close! Thank you for all the work!

PR #41, opened 2023-01-19
Decode single message for a conversation

nakajima:
Also adds the ability to serialize a `Conversation` via `Conversation.encodedContainer`.

See https://github.com/xmtp/xmtp-ios/issues/31.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ let keysData = try keys.serializedData()
 Once you have those keys, you can create a new client with `Client.from`:
 
 ```swift
+let keys = try PrivateKeyBundleV1(serializedData: keysData)
 let client = try Client.from(bundle: keys, options: .init(api: .init(env: .production)))
 ```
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -253,6 +254,43 @@ let myAppConversations = conversations.filter {
...
 
+### Decoding a single message
+
+You can decode a single `Envelope` from XMTP using the `decode` method:
+
+```swift
+let conversation = try await client.conversations.newConversation(with: "0x3F11b27F323b62B159D2642964fa27C46C841897")
+
+// Assume this function returns an Envelope that contains a message for the above conversation
+let envelope = getEnvelopeFromXMTP()
```
elisealix22, line 265: Out of curiosity, can you get an `Envelope` from a `DecodedMessage` or is there another way to do it?

elisealix22 (approved):
🔥

README.md
nakajima: I think my understanding was that you'd get an `Envelope` from a push notification potentially, then need to decode it with this? If you've already got a `DecodedMessage` I'm not sure what you'd need to decode at that point.

PR #42, opened 2023-01-19
Enable message compression/decompression

nakajima:
Adds the ability to send and receive compressed messages [as described by the JS SDK](https://github.com/xmtp/xmtp-js#compression).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,31 +14,31 @@ jobs:
         with:
           xcode-version: latest-stable
       - uses: actions/checkout@v3
-      - name: Update Homebrew and save docker version
-        run: |
-          brew update --preinstall
-          cat "$(brew --repository)/Library/Taps/homebrew/homebrew-core/Formula/docker.rb" > .github/brew-formulae
-      - name: Configure Homebrew docker cache files
-        uses: actions/cache@v3
-        with:
-          path: |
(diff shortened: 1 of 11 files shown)


.github/workflows/test.yml
```diff
@@ -14,31 +14,31 @@ jobs:
...
-          # Docker engine is no longer available because of licensing
-          # Alternative Colima is part of the github macOS runner
-          colima start
-          # Ensure colima is configured for later user
-          echo "DOCKER_HOST=unix://$HOME/.colima/default/docker.sock" >> $GITHUB_ENV
-          # Verify Docker
-          docker ps
-      - name: Start local test server
-        run: docker-compose -p xmtp -f dev/local/docker-compose.yml up -d
+      # - name: Update Homebrew and save docker version
```
elisealix22, line 17: Any reason this is commented out now?

elisealix22 (approved):

.github/workflows/test.yml
nakajima: It was causing some test failures (possibly related to https://github.com/xmtp/xmtp-js/pull/262?). I wanted to comment it out to see if that fixed things and it did. I think it'd be good to be able to run a lot of the integration tests in CI at some point so it's probably worth keeping that stuff around and commented, but it's not strictly necessary at this moment.

PR #43, opened 2023-01-21
Add Push Notification example

nakajima:
Add example that uses https://github.com/xmtp/example-notification-server-go to receive push notifications.


nakajima:
I wonder if _all_ of this should be in the example app actually instead of adding certain parts to the SDK. It feels like it'd be nice to have more batteries included for the SDK but also maybe this is too presumptive?

nplasterer (approved):
I think this seems fine to include. But I'll lean on @elisealix22 for her thoughts since there maybe be better places to show this as an example without making this base example too complicated.

elisealix22 (approved):
The push abstraction makes sense to me! I like that the persistence layer is still in the example app for now too. I ended up following these two examples for storing keys in the 1P app so keeping that flexible makes sense:
1. https://developer.apple.com/documentation/cryptokit/storing_cryptokit_keys_in_the_keychain
2. https://johncodeos.com/how-to-save-sensitive-data-in-keychain/

Sources/XMTP/Push/XMTPPush.swift
```diff
@@ -0,0 +1,90 @@
+//
```
neekolas, line 1: What are the implications of having this live outside of the example folder?

It seems like if we want to treat this as a library, we are going to have to support it as such (documentation, API stability)

I wonder if it makes sense to move this into the example folder for now to lower expectations. Genuinely unsure of the right answer.

I'm not enough of a Swift expert to understand how much work that's going to put on the user to copy/paste/regenerate.

XMTPiOSExample/XMTPiOSExample/Views/ConversationListView.swift
```diff
@@ -66,10 +69,29 @@ struct ConversationListView: View {
...
+			await add(conversations: conversations)
 		} catch {
 			print("Error loading conversations: \(error)")
 		}
 	}
+
+	func add(conversations: [Conversation]) async {
+		// Ensure we're subscribed to push notifications on these conversations
+		do {
+			try await XMTPPush.shared.subscribe(topics: conversations.map(\.topic))
```
neekolas, line 82: For a full featured push notification example, we are going to have to subscribe to more than just conversation topics.

We are going to want to subscribe to the `invite` and `intro` topics to trigger a background update to the conversation list when a new invite/intro is received.

Sources/XMTP/Push/XMTPPush.swift
```diff
@@ -0,0 +1,90 @@
...
+			request.deliveryMechanism = Notifications_V1_DeliveryMechanism.with { delivery in
+				delivery.apnsDeviceToken = token
+				delivery.deliveryMechanismType = .apnsDeviceToken(token)
+			}
+		}
+
+		_ = await client.registerInstallation(request: request)
+	}
+
+	public func subscribe(topics: [String]) async throws {
```
neekolas, line 67: If we did want this to function more as a library that we expect other devs to use directly, I would expect this to include a `revoke` method that calls `deleteInstallation` as well as a way to `unsubscribe`.

Although for right now, neither of those are really needed to build a basic push example.

Sources/XMTP/Push/XMTPPush.swift
nakajima: Having it over here means it's part of the SDK and anyone can just use it.

> I wonder if it makes sense to move this into the example folder for now to lower expectations. Genuinely unsure of the right answer.

Ya this is the biggest I have as well. I really like the idea that adding push to your app is as simple as using `XMTPPush`, but it does make a hard assumption that you're running [xmtp/example-notification-server-go](https://github.com/xmtp/example-notification-server-go) somewhere.

I also like the flexibility that anyone can write any software that subscribes to the network and delivers pushes in any format they want, I just wonder if that's a lot to be asking for folks who just want to get something working, especially iOS devs who may not really know much about setting up servers.

XMTPiOSExample/XMTPiOSExample/Views/ConversationListView.swift
nakajima: That's a good call, I should add it to this PR.

Sources/XMTP/Push/XMTPPush.swift
nakajima: Since I'm leaning more toward the library approach, I think this would be good to add.

nakajima:
Gonna merge this as is just because it's been open for a while. We can explore more of the line between library vs example in follow ups.

neekolas:
👍🏼

PR #45, opened 2023-01-21
Make SendOptions work more like the JS SDK

nakajima:
This brings sending behavior more in line with what the JS SDK does. Also some README updates for things that were out of date.


nplasterer (approved):
LGTM. I'm going to have to change some things I've done in Kotlin.

PR #62, opened 2023-02-14
Attachments proposal

nakajima:
An idea for how to handle media/file attachments. Would definitely need some more validations around size and whatnot.


Sources/XMTP/Codecs/AttachmentCodec.swift
```diff
@@ -0,0 +1,57 @@
...
+	public var filename: String
+	public var mimeType: String
+	public var data: Data
+
+	public init(filename: String, mimeType: String, data: Data) {
+		self.filename = filename
+		self.mimeType = mimeType
+		self.data = data
+	}
+}
```
nakajima, line 27: This is what the encoded content looks like.

nakajima:
Closing in favor of https://github.com/xmtp/xmtp-ios/pull/62.

PR #68, opened 2023-02-20
Implement remote attachments

nakajima:
Implements [Remote Attachments](https://github.com/xmtp/XIPs/blob/624890e12391755095a4517b64728fd29a00c68f/XIPs/xip-17-remote-attachment-content-type-proposal.md).


elisealix22 (approved):
Incredible to see this land.

Also /cc @nplasterer since we'll likely want this in the Android SDK as well! 👀

Sources/XMTP/Client.swift
```diff
@@ -63,6 +63,8 @@ public class Client {
 	static var codecRegistry = {
 		var registry = CodecRegistry()
 		registry.register(codec: TextCodec())
+		registry.register(codec: AttachmentCodec())
```
neekolas, line 66: Missed this in the first pass, but I don't think we should default to enabling any additional codecs in the client. A new app developer should be able to make something that just supports text, and rely on the fallback of the content type to handle more complex things like attachments and remote attachments.

nakajima: I can get down with that. PR in https://github.com/xmtp/xmtp-ios/pull/72.

PR #72, opened 2023-03-01
Opt into attachment codecs

nakajima:
See https://github.com/xmtp/xmtp-ios/pull/68#discussion_r1121019792


neekolas (approved):
Thanks for making this change so quick

PR #75, opened 2023-03-06
Implement streamAllMessages

nakajima:
Also includes some formatting fixes.

I removed the notes from the README about this not being supported, but the README could definitely use a bigger refresh to reflect all the changes that have happened recently.


nplasterer (approved):
Probably going to want this in kotlin as well. Going to take a stab.

PR #77, opened 2023-03-08
Ensure pre-key signed by identity key

nakajima:
See https://github.com/xmtp/xmtp-js/pull/298.

### DIFF
--- a/Sources/XMTP/Messages/MessageV2.swift
+++ b/Sources/XMTP/Messages/MessageV2.swift
@@ -34,6 +34,8 @@ extension MessageV2 {
 			let senderPreKey = try PublicKey(signed.sender.preKey)
 			let senderIdentityKey = try PublicKey(signed.sender.identityKey)
 
+			// This is a bit confusing since we're passing keyBytes as the digest instead of a SHA256 hash.
+			// That's because our underlying crypto library always SHA256's whatever data is sent to it for this.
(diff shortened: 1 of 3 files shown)


nplasterer (approved):

Sources/XMTP/Messages/MessageV2.swift
```diff
@@ -27,6 +27,17 @@ extension MessageV2 {
...
 			let signed = try SignedContent(serializedData: decrypted)
 
+			guard signed.sender.hasPreKey, signed.sender.hasIdentityKey else {
+				throw MessageV2Error.decodeError("missing sender pre-key or identity key")
+			}
+
+			let senderPreKey = try PublicKey(signed.sender.preKey)
+			let senderIdentityKey = try PublicKey(signed.sender.identityKey)
+
+			if !(try senderPreKey.signature.verify(signedBy: senderIdentityKey, digest: signed.sender.preKey.keyBytes)) {
```
nakajima, line 37: This is a bit confusing since we're passing `keyBytes` as the `digest` instead of a SHA256 hash. That's because our underlying crypto library always SHA256's whatever data is sent to it for this.

nakajima: I'm actually going to add this a comment in the code because this isn't the first time this has bit me.

michaelx11 (approved):
LGTM once tests pass! Let me know if/when it ships and I can test it manually too

PR #79, opened 2023-03-09
standardize clientVersion string

jazzz:
## Problem

The default version string does not contain a reference to xmtp-ios, which is not the standard across xmtp sdks

## Solution

Standardize this client string to `xmtp-ios/0.0.0-development` so it matches the other SDK entries.

## Open Questions

- Was this intentional? Could be I am missing some context here.


nakajima (approved):
This wasn't intentional, good catch.

PR #80, opened 2023-03-09
Introduce Conversation.prepareMessage

nakajima:
This gives back a PreparedMessage that has a messageID on it. Clients can use this for optimistic sending purposes. Clients can then call `send()` on the prepared message to publish it to the network.

### DIFF
--- a/Sources/XMTP/PreparedMessage.swift
+++ b/Sources/XMTP/PreparedMessage.swift
@@ -1,0 +1,27 @@
+//
+//  PreparedMessage.swift
+//
+//
+//  Created by Pat Nakajima on 3/9/23.
+//
+
+import CryptoKit
+import Foundation
+
+public struct PreparedMessage {
+	var messageEnvelope: Envelope
+	var conversation: Conversation
+	var onSend: () async throws -> Void
+
+	public func decodedMessage() throws -> DecodedMessage {
+		return try conversation.decode(messageEnvelope)
+	}
+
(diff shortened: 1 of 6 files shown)


Sources/XMTP/PreparedMessage.swift
```diff
@@ -0,0 +1,27 @@
...
+import CryptoKit
+import Foundation
+
+public struct PreparedMessage {
+	var messageEnvelope: Envelope
+	var conversation: Conversation
+	var onSend: () async throws -> Void
+
+	public func decodedMessage() throws -> DecodedMessage {
+		return try conversation.decode(messageEnvelope)
```
nakajima, line 17: Could maybe have prepared message include EncodedContent and stuff so we don't need to involve the encryption stuff here, but this seems like an OK start for now.

nplasterer (approved):
This looks good to me. Was just wondering if we should try to keep this stuff consistent between flutter, ios, and android or if we don't care as much anymore. I just saw this flutter PR come through this morning doing something sort of similar for performance around message sending and encoding. https://github.com/xmtp/xmtp-flutter/pull/55/files

nakajima:
It probably makes sense to standardize on the ability to create a message locally for sending across SDKs, we should probably use similar terminology even if the APIs aren’t the exact same. I hadn’t seen the flutter pr but I’ll take a look to see if we can unify a bit.

PR #81, opened 2023-03-14
Remove multiple topic queries

nakajima:
In a move towards decentralization we need to remove the ability to query multiple topics as we can't guarantee topics will be on the same node.

See also https://github.com/xmtp/xmtp-android/pull/52.


nplasterer (approved):

mkobetic (approved):
LGTM, thank you! ❤️

PR #93, opened 2023-04-20
Use Buf to generate proto code

neekolas:
## Summary

- Adds automated generation of proto files
-  Removes`message_api.grpc.swift` entirely, instead of hiding the code using `#if canImport(GRPC)`
- Adds a script to regenerate protos (has to be run manually right now)

## Notes
Other than the removal of `message_api.grpc.swift` it matches the existing files in the Proto folder exactly (I deleted the folder, re-ran, and these are the only diffs)


nakajima (approved):

michaelx11 (approved):
This takes care of the gRPC dependency without needing to vendor proto files or use a custom branch right? @neekolas - if so ship it 🚢

neekolas:
At least right now, it's just a better way to vendor proto files and keep them up to date

PR #91, opened 2023-04-18
Cocoapods integration

nakajima:
## Overview

Issue: https://github.com/xmtp/xmtp-react-native-old/issues/10

This PR represents the changes needed to package xmtp-ios into an XMTP cocoapod.

The major changes are:
1) Replace references to 'xmtp/proto' with buf #93 
2) Use the XMTPRust cocoapod to replace gRPC and other non-cocoapod-compatible code such as some secp256k1 operations ([xmtp-rust-swift](https://github.com/xmtp/xmtp-rust-swift))
3) Change ApiClient and other Swift classes to accommodate XMTPRust, specifically in KeyUtil.swift and for signature verification logic

## Diagram

See https://github.com/xmtp/xmtp-rust-swift#diagram 

## Releases

From this branch we've released:

- 0.1.3-beta0 - this is the first XMTP cocoapod release, targeting XMTPRust 0.1.2-beta0 which did unnecessary JSON serialization in the Rust<>Swift boundary and had broken streaming (only got first item). Pagination info not returned by API.
- 0.2.0-alpha0 - do not use, this release did not properly handle Pagination info from `query` and therefore only returns the first 100 elements of a query
- 0.2.1-alpha0 - targets XMTPRust 0.2.2-beta0 - Does not use serialization in Rust<>Swift boundary, instead converts between RustVec<UInt8> and Data. Returns pagination info in `query` and has working streaming.

## Tests

- Ensure all unit tests pass (with a local node)
- Added new PaginationTests which exercise Pagination, one of the trickier parts of the ApiClient migration to Rust code
- XCode cloud keeps failing, seems related to misconfigured XMTP scheme

PR #97, opened 2023-04-24
Integrate xmtp-rust-swift 0.2.0-beta0

nakajima:
Updates Swift side for changes in https://github.com/xmtp/xmtp-rust-swift/pull/6.

For followups: 

- Look at subscription implementation
- Wrap more of the Rust types in a more Swifty interface


Sources/XMTP/ApiClient.swift
```diff
@@ -105,9 +146,22 @@ class GRPCApiClient: ApiClient {
...
 		return AsyncThrowingStream { continuation in
 			Task {
-                for try await envelopeJson in self.rustClient.subscribe(topics: topics) {
-                    let envelope = try Envelope(jsonString: envelopeJson)
-					continuation.yield(envelope)
+				let topicsVec = RustVec<RustString>()
+				for topic in topics {
+					topicsVec.push(value: topic.intoRustString())
+				}
+				let subscription = try await self.rustClient.subscribe(topicsVec)
```
neekolas, line 153: One thing worth calling out is that I don't do any sort of retries in the Rust layer. Drive through a tunnel, deploy the nodes, whatever: the subscription will end. I've been assuming the caller would handle retries, but if that's not viable we can move the logic into Rust.

elisealix22 (approved):
Thank goodness for tests ✅

Sources/XMTP/ApiClient.swift
```diff
@@ -71,14 +106,20 @@ class GRPCApiClient: ApiClient {
...
-        var encodedPaging = ""
-        if request.hasPagingInfo {
-            encodedPaging = try request.pagingInfo.jsonString()
-        }
-        let responseJson = try await rustClient.query(topic: topic, json_paging_info: encodedPaging)
-        // Decode the response as a QueryResponse
-        let decodedQueryResponse = try QueryResponse(jsonString: responseJson)
-        return decodedQueryResponse
+		let paging = XMTPRust.PagingInfo(limit: 0, cursor: nil, direction: XMTPRust.SortDirection.Ascending)
+		let response = try await rustClient.query(topic.intoRustString(), Optional.none, Optional.none, Optional.none)
```
neekolas, line 110: Shouldn't we be passing in the `pagingInfo`, `startTime`, and `endTime` here?

Sources/XMTP/ApiClient.swift
```diff
@@ -71,14 +106,20 @@ class GRPCApiClient: ApiClient {
...
+		// Build the query response from response fields
+		queryResponse.envelopes = response.envelopes().map { rustEnvelope in
+			var envelope = Envelope()
+			envelope.contentTopic = rustEnvelope.get_topic().toString()
+			envelope.timestampNs = rustEnvelope.get_sender_time_ns()
+			envelope.message = dataFromRustVec(rustVec: rustEnvelope.get_payload())
+			return envelope
+		}
+		// Decode the response as a QueryResponse
+		return queryResponse
```
neekolas, line 122: The response paging info seems to be getting lost here.

Sources/XMTP/ApiClient.swift
michaelx11: good catch!

michaelx11:
I think the xcode cloud tests are broken on the base branch (cocoapods-integration). Tried a bit to tweak it but this no-op [PR](https://github.com/xmtp/xmtp-ios/pull/99) seems to show it's just broken.

Happy to merge this into cocoapods-integration then we can figure out how to get the tests pass on that branch in it's current PR #91

michaelx11:
@nakajima - since the xcode cloud tests are broken on `cocoapods-integration` base [test PR](https://github.com/xmtp/xmtp-ios/pull/99), I'm going to merge this PR into `cocoapods-integration` and then we can open a subsequent PR to fix it.

I'm worried that as time goes on, we lose our chance to add 0.2.0-beta0 support (aka working subscribe functionality)

michaelx11:
Oh wait I fixed it :)

PR #100, opened 2023-05-10
fix: with Nick's PR we can now remove the xmtp/proto swift dependency

michaelx11:
## Overview

Following #93 getting merged in, we don't have to rely on a special-cased xmtp/proto branch that doesn't have gRPC code anymore.

## Test Plan

- Unit tests still work
- Will verify that xcode cloud Github actions pass


michaelx11:
Hmm, I still don't have a good read on why the XMTPIosExample Test action is failing. I don't have access to Xcode Cloud either :/

PR #101, opened 2023-05-11
fix: remove proto dependency from Package.swift

michaelx11:
## Overview

Following #93 getting merged in, we don't have to rely on a special-cased xmtp/proto branch that doesn't have gRPC code anymore.

## Test Plan

- Unit tests still work
- Will verify that xcode cloud Github actions pass


michaelx11:
Running `pod install` in the example directory seems to have fixed the issue

neekolas (approved):

PR #102, opened 2023-05-12
fix: local tests failing due to improper mem copying, .bytes extension not available in pods

michaelx11:
## Overview

Two issues were preventing the cocoapods-integration branch from passing `pod lib lint XMTP.podspec  --allow-warnings`, the precursor to actually shipping the cocoapod again.

1) Indirect problem: we were utilizing subranges of Data when storing the result of Swift AES.GCM.Seal, this meant in a subsequent index like `payload[0]` we would hit a crash. Specifically, let `output = AES.GCM.Seal(...)` then `output.ciphertext` is actually a (12, ...) range of a shared underlying Data instance (`.combined`). We use `subdata` to create a copy to allow for safer indexing going forward.

NOTE: it seems possible to create a new Data representation over the same memory location with different offset, but that can be an optimization. Copying memory here isn't the worst as it's outweighed by the encryption operation and the eventual serialization.

2) Direct problem: We worked around this by using `.bytes` Data extension, which returned an unsafe pointer over the Data (see Note above). This works but was an extension provided by the `sec256k1` Swift package that is excluded from our podspec because it's not Cocoapods compliant.

## Test Plan

- Unit tests pass
- Passes pod lint

### DIFF
--- a/Sources/XMTP/Crypto.swift
+++ b/Sources/XMTP/Crypto.swift
@@ -35,8 +35,14 @@ enum Crypto {
 		}
 
 		var ciphertext = CipherText()
-
-		ciphertext.aes256GcmHkdfSha256.payload = payload.ciphertext + payload.tag
+		// Copy the ciphertext data out, otherwise it's a region sliced from a combined Data (nonce, ciphertext, tag)
+		// with offsets like lowerBound=12, upperBound=224. Without copying, trying to index like payload[0] crashes
+		// up until payload[12]. This is mostly a problem for unit tests where we decrypt what we encrypt in memory, as
+		// serialization/deserialization acts as copying and avoids this issue.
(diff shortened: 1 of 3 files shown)


Sources/XMTP/Crypto.swift
```diff
@@ -35,8 +35,14 @@ enum Crypto {
 		}
 
 		var ciphertext = CipherText()
-
-		ciphertext.aes256GcmHkdfSha256.payload = payload.ciphertext + payload.tag
```
michaelx11, line 39: This code left us with `ciphertext.aes256GcmHkdfSha256.payload` being composed of Data subscripts: https://forums.swift.org/t/data-subscript/57195

So subsequent operations like `ciphertext.aes256GcmHkdfSha256.payload[0]` would crash with EXC_BAD_INSTRUCTION

michaelx11:
FYI @nakajima gonna send this

PR #108, opened 2023-06-12
feat: mem cache convo list, fix convo id, add error types

dmccartney:
This adds memory caching to the `conversations.list()` method -- after the first call, later calls will only look for conversations _newer_ than the most recently listed.

This also got the integration tests running with an easier toggle to enable running against a local node.
And this fixes bugs that cropped up trying to get tests to pass (e.g. `newConversation` now properly detects new/old conversations).


Sources/XMTPTestHelpers/TestHelpers.swift
```diff
@@ -11,6 +11,35 @@ import XCTest
...
+    // static let TEST_SERVER_HOST = _env("TEST_SERVER_HOST") ?? "127.0.0.1"
+    // static let TEST_SERVER_PORT = Int(_env("TEST_SERVER_PORT")) ?? 5556
+    // static let TEST_SERVER_IS_SECURE = _env("TEST_SERVER_IS_SECURE") == "true"
+
+    static private func _env(_ key: String) -> String? {
+        ProcessInfo.processInfo.environment[key]
+    }
+
+    static public func skipIfNotRunningLocalNodeTests() throws {
+        try XCTSkipIf(!TEST_SERVER_ENABLED, "requires local node")
```
dmccartney, line 26: This makes it so that you can run all local-node tests via:
```
$ TEST_SERVER_ENABLED=true swift test
```
(and in CI, once we get the server running, we can also pass in the `env:` to enable them)

nplasterer (approved):
LGTM. But might be worth getting eyes from @richardhuaaa on the Rust stuff.

PR #111, opened 2023-06-21
Add ability to send EncodedContent objects directly

nakajima:
For React Native, we want to be able to create `EncodedContent` on the react native side then serialize and send those to native land to be actually encrypted and sent to the network. This change enables that.


daria-github (approved):

nplasterer:
@daria-github I don't think you can use this in react native until you bump the podspec, tag it, and publish it.

Issue #86, opened 2023-03-29
Implement RAPID to dedupe conversations in iOS


yash-luna:
Based on JS implementation, https://github.com/xmtp/xmtp-js/issues/316, after it's complete

PR #115, opened 2023-06-29
feat: deterministic topic/keyMaterial for invites

dmccartney:
This produces consistent `topic` and `keyMaterial` for invites created to the same `peer` and `conversationId`.

This implements #86

### DIFF
--- a/Sources/XMTP/Messages/Invitation.swift
+++ b/Sources/XMTP/Messages/Invitation.swift
@@ -1,9 +1,6 @@
 //
 //  Invitation.swift
 //
-//
-//  Created by Pat Nakajima on 11/26/22.
-//
 
 import Foundation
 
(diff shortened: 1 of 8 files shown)


Sources/XMTP/Messages/Invitation.swift
```diff
@@ -1,35 +1,44 @@
...
-//  Created by Pat Nakajima on 11/26/22.
-//
 
 import Foundation
 
 /// Handles topic generation for conversations.
 public typealias InvitationV1 = Xmtp_MessageContents_InvitationV1
 
 extension InvitationV1 {
-	static func createRandom(context: InvitationV1.Context? = nil) throws -> InvitationV1 {
```
nplasterer, line 14: @nakajima and I talked about there still being a need for `createRandom` And so `createDeterministic` wouldn't necessarily remove it but we can always add it back.

nplasterer (approved):

PR #143, opened 2023-08-04
fix: use type from the reply nested content directly

dmccartney:
This gets the nested type of a reply from the deserialized `EncodedContent` instead of inspecting the parameter map.


nplasterer (approved):
We will need to bump the pod spec in here before I can publish it for react native.

PR #148, opened 2023-08-23
feat: include topic in the DecodedMessage

dmccartney:
This adds the conversation `topic` to the `DecodedMessage` so that it can be associated with the corresponding conversation.

This is part of fix for the leftover bits from https://github.com/xmtp/xmtp-react-native/pull/86

This mirrors the Android edition from https://github.com/xmtp/xmtp-android/pull/110


nplasterer (approved):

nplasterer:
Since we need this in react native we should bump the pod spec in here so we can cut a new pod release.

PR #151, opened 2023-09-05
fix: normalize reaction encoding, support legacy form

dmccartney:
This fixes the iOS portion of how reactions are encoded across codec implementations.

See https://github.com/xmtp/xmtp-react-native/issues/93#issuecomment-1707220192
See also https://github.com/xmtp/xmtp-js-content-types/pull/26


nplasterer (approved):
Mind bumping the pod spec in this PR as well so we can get it in react native?

PR #157, opened 2023-09-15
Remove timestamp from read receipts

nplasterer:
Implements https://github.com/orgs/xmtp/discussions/43#discussioncomment-6972516

This removes the timestamp from the read receipt content type in favor of the timestamp already on the message.

### DIFF
--- a/Sources/XMTP/Codecs/ReadReceiptCodec.swift
+++ b/Sources/XMTP/Codecs/ReadReceiptCodec.swift
@@ -9,13 +9,7 @@ import Foundation
 
 public let ContentTypeReadReceipt = ContentTypeID(authorityID: "xmtp.org", typeID: "readReceipt", versionMajor: 1, versionMinor: 0)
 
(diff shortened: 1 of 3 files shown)


Sources/XMTP/Codecs/ReadReceiptCodec.swift
```diff
@@ -9,13 +9,7 @@ import Foundation
...
 public let ContentTypeReadReceipt = ContentTypeID(authorityID: "xmtp.org", typeID: "readReceipt", versionMajor: 1, versionMinor: 0)
 
-public struct ReadReceipt {
-    public var timestamp: String
-    
-    public init(timestamp: String) {
-        self.timestamp = timestamp
-    }
-}
+public struct ReadReceipt {}
```
nakajima, line 12: Not sure we need a struct for this. Maybe we just set the `T` to `Never`?

nakajima (approved):

Issue #159, opened 2023-09-15
Bug: Framework can cause app to crash due to variable data race with async access


benjaminmayo:
### Describe the bug

The internal `conversationsByTopic` property can be accessed and mutated from various threads simultaneously due to concurrency. As access to this storage is not gated by a lock or other synchronization, this can lead to malformed memory and cause unavoidable crashes when the framework accesses that property. 

See screenshot for an example of where Xcode caught the error in the debugger:
<img width="1840" alt="Screenshot 2023-09-15 at 23 44 57" src="https://github.com/xmtp/xmtp-ios/assets/713431/43db74d5-a972-479a-9725-17bb717d9e62">

### Expected behavior

The XMTP framework should not induce crashes. 

### Steps to reproduce the bug

Client code that invokes Conversations methods which internally mutate the `conversationsByTopic` storage will eventually trigger the crash. This was first observed due to app code that observed `streamAllMessages()` and then also invoked the `list()` method.

neekolas:
Thanks for flagging this. Great catch!

PR #160, opened 2023-09-17
Fix conversationsByTopic concurrency issues

nplasterer:
Fixes https://github.com/xmtp/xmtp-ios/issues/159

Use async and await to fix concurrency issues around `conversationsByTopic` that was causing crashes.

### DIFF
--- a/Sources/XMTP/Conversations.swift
+++ b/Sources/XMTP/Conversations.swift
@@ -7,7 +7,7 @@ public enum ConversationError: Error {
 /// Handles listing and creating Conversations.
 public class Conversations {
     var client: Client
-    var conversationsByTopic: [String: Conversation] = [:]
+    @MainActor var conversationsByTopic: [String: Conversation] = [:]
 
     init(client: Client) {
         self.client = client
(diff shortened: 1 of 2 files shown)


Sources/XMTP/Conversations.swift
```diff
@@ -79,16 +81,20 @@ public class Conversations {
...
                     do {
                         for try await envelope in client.subscribe(topics: topics) {
-                            if let conversation = conversationsByTopic[envelope.contentTopic] {
+                            if let conversation = await conversationsByTopic[envelope.contentTopic] {
                                 let decoded = try conversation.decode(envelope)
                                 continuation.yield(decoded)
                             } else if envelope.contentTopic.hasPrefix("/xmtp/0/invite-") {
                                 let conversation = try fromInvite(envelope: envelope)
-                                conversationsByTopic[conversation.topic] = conversation
+                                async {
```
benjaminmayo, line 89: I think these should be `await`s, not async tasks. If you make a task here, the loop can break before the dictionary value is written, so it will miss the subscription.

PR #161, opened 2023-09-19
Make Conversations and Contacts actors.

nakajima:
Hopefully will fix https://github.com/xmtp/xmtp-ios/issues/159.

This PR also removes some failing v1 tests that are no longer applicable.


nplasterer (approved):
Can you bump the pod spec

PR #174, opened 2023-10-11
Bump Rust Version

nplasterer:
This bumps to the latest Rust version which drops the timeouts lower so they aren't noticeable.

I was not able to reproduce `Remote attachments don’t work between the xmtp.chat and iOS XMTP.RemoteAttachmentError.invalidScheme("invalid scheme value. must be https://")` But I modified some of the error messages to confirm that the latest version of swift is getting run.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -26,7 +26,7 @@ let package = Package(
 		.package(url: "https://github.com/1024jp/GzipSwift", from: "5.2.0"),
 		.package(url: "https://github.com/bufbuild/connect-swift", from: "0.3.0"),
 		.package(url: "https://github.com/apple/swift-docc-plugin.git", from: "1.0.0"),
-		.package(url: "https://github.com/xmtp/xmtp-rust-swift", from: "0.3.1-beta0"),
+		.package(url: "https://github.com/xmtp/xmtp-rust-swift", from: "0.3.5-beta0"),
 	],
 	targets: [
 		// Targets are the basic building blocks of a package. A target can define a module or a test suite.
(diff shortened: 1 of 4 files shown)


Sources/XMTP/Codecs/RemoteAttachmentCodec.swift
```diff
@@ -74,7 +74,7 @@ public struct RemoteAttachment {
 
 	func ensureSchemeMatches() throws {
 		if !url.hasPrefix(scheme.rawValue) {
-			throw RemoteAttachmentError.invalidScheme("scheme must be https://")
+			throw RemoteAttachmentError.invalidScheme("scheme must be https")
 		}
 	}
```
nplasterer, line 79: @nakajima I think this is where the error is getting thrown. Is there a reason we call this on every init?

Sources/XMTP/Codecs/RemoteAttachmentCodec.swift
```diff
@@ -174,7 +174,7 @@ public struct RemoteAttachmentCodec: ContentCodec {
 			throw RemoteAttachmentError.invalidScheme("no scheme parameter")
 		}
 
-        if (!schemeString.starts(with: "https")) {
+        if (!schemeString.hasPrefix(RemoteAttachment.Scheme.https.rawValue)) {
```
nplasterer, line 177: Changed this to hasPrefix since that seems to be more inline with how swift would do this.

Package.swift
```diff
@@ -26,7 +26,7 @@ let package = Package(
 		.package(url: "https://github.com/1024jp/GzipSwift", from: "5.2.0"),
 		.package(url: "https://github.com/bufbuild/connect-swift", from: "0.3.0"),
 		.package(url: "https://github.com/apple/swift-docc-plugin.git", from: "1.0.0"),
-		.package(url: "https://github.com/xmtp/xmtp-rust-swift", from: "0.3.1-beta0"),
+		.package(url: "https://github.com/xmtp/xmtp-rust-swift", from: "0.3.5-beta0"),
```
nplasterer, line 29: Bumping this for anyone using SPM as well.

nakajima (approved):

PR #189, opened 2023-11-14
Allow envelopes to be decrypted but not decoded

nakajima:
This will give us flexibility to handle more content types in React Native without double decoding.


langer:
Extremely off-topic comment here but just cool and vaguely mind-blowing to open Github and randomly see an old NYC friend and an Instagram friend in conversation on a pull request.

PR #193, opened 2023-11-24
Bump the pod

nplasterer:
The pod spec is failing with `[!] Unable to accept duplicate entry for: XMTP (0.6.8-alpha0)`

https://github.com/xmtp/xmtp-ios/actions/runs/6982329393

Needed so we can fix this https://github.com/xmtp/xmtp-react-native/issues/153


kele-leanes (approved):
LGTM

PR #194, opened 2023-11-25
Removing conversations that contain a no valid topic

giovasdistillery:
### Expectations
- Adding a filter to not show conversations that contain no valid topics
- The filter for now is allowing topics with no special characters

### DIFF
--- a/Sources/XMTP/Conversations.swift
+++ b/Sources/XMTP/Conversations.swift
@@ -248,9 +248,8 @@ public actor Conversations {
     public func isValidTopic(topic: String) -> Bool {
         let regex = "^[\\x00-\\x7F]+$"
         let initialIndex = String.Index(utf16Offset: 8, in: topic)
-        let finalIndex = topic.lastIndex(of: "/")
-        if finalIndex != nil {
-            let unwrappedTopic = String(topic[initialIndex..<finalIndex!])
+        if let finalIndex = topic.lastIndex(of: "/") { 
+            let unwrappedTopic = String(topic[initialIndex..<finalIndex])
             let indices = unwrappedTopic.startIndex..<unwrappedTopic.endIndex
             let resultMatch = unwrappedTopic.range(of: regex, options: .regularExpression)
             return indices == resultMatch
--- a/Sources/XMTP/Conversations.swift
+++ b/Sources/XMTP/Conversations.swift
@@ -236,13 +236,26 @@ public actor Conversations {
         }
 
         newConversations
-            .filter { $0.peerAddress != client.address }
+            .filter { $0.peerAddress != client.address && isValidTopic(topic: $0.topic) }
             .forEach { conversationsByTopic[$0.topic] = $0 }
 
         // TODO(perf): use DB to persist + sort
         return conversationsByTopic.values.sorted { a, b in
             a.createdAt < b.createdAt
         }
+    }
+    
+    public func isValidTopic(topic: String) -> Bool {
+        let regex = "^[\\x00-\\x7F]+$"
+        let initialIndex = String.Index(utf16Offset: 8, in: topic)
+        let finalIndex = topic.lastIndex(of: "/")
+        if finalIndex != nil {
+            let unwrappedTopic = String(topic[initialIndex..<finalIndex!])
+            let indices = unwrappedTopic.startIndex..<unwrappedTopic.endIndex
+            let resultMatch = unwrappedTopic.range(of: regex, options: .regularExpression)
+            return indices == resultMatch
+        }
+        return false
     }
 
     private func listIntroductionPeers(pagination: Pagination?) async throws -> [String: Date] {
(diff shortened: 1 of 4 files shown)


Sources/XMTP/Conversations.swift
```diff
@@ -236,14 +236,27 @@ public actor Conversations {
...
+    public func isValidTopic(topic: String) -> Bool {
+        let regex = "^[\\x00-\\x7F]+$"
+        let initialIndex = String.Index(utf16Offset: 8, in: topic)
+        let finalIndex = topic.lastIndex(of: "/")
+        if finalIndex != nil {
+            let unwrappedTopic = String(topic[initialIndex..<finalIndex!])
+            let indices = unwrappedTopic.startIndex..<unwrappedTopic.endIndex
+            let resultMatch = unwrappedTopic.range(of: regex, options: .regularExpression)
+            return indices == resultMatch
+        }
```
nplasterer, line 257: I think this is why the linter is failing with `error: Force Unwrapping Violation: Force unwrapping should be avoided (force_unwrapping)` I also don't think this is the most swifty way to do this but @nakajima might now better.

```suggestion
        if let finalIndex = topic.lastIndex(of: "/") { 
            let unwrappedTopic = String(topic[initialIndex..<finalIndex])
            let indices = unwrappedTopic.startIndex..<unwrappedTopic.endIndex
            let resultMatch = unwrappedTopic.range(of: regex, options: .regularExpression)
            return indices == resultMatch
        }
```

nplasterer (commented):
Do you mind bumping the podfile like what I did here: https://github.com/xmtp/xmtp-ios/pull/193 so we can get these changes into react native as well.

Also looks like a legit lint error.

Sources/XMTP/Conversations.swift
```diff
@@ -236,14 +236,26 @@ public actor Conversations {
...
+    public func isValidTopic(topic: String) -> Bool {
+        let regex = "^[\\x00-\\x7F]+$"
+        let initialIndex = String.Index(utf16Offset: 8, in: topic)
+        if let finalIndex = topic.lastIndex(of: "/") { 
+            let unwrappedTopic = String(topic[initialIndex..<finalIndex])
+            let indices = unwrappedTopic.startIndex..<unwrappedTopic.endIndex
+            let resultMatch = unwrappedTopic.range(of: regex, options: .regularExpression)
+            return indices == resultMatch
+        }
+        return false
```
nakajima, line 257: Could we maybe just check to see if the string is all valid ASCII? Something like this:

```suggestion
        topic.allSatisfy(\.isASCII)
```

Also, it might make more sense to make the validation a static func on `Topic`?

Tests/XMTPTests/ConversationsTest.swift
```diff
@@ -78,4 +78,64 @@ class ConversationsTests: XCTestCase {
...
+        
+        // Creation of all known types of topics
+        let privateStore = Topic.userPrivateStoreKeyBundle(validId).description
+        let contact = Topic.contact(validId).description
+        let userIntro = Topic.userIntro(validId).description
+        let userInvite = Topic.userInvite(validId).description
+        let directMessageV1 = Topic.directMessageV1(validId, "sd").description
+        let directMessageV2 = Topic.directMessageV2(validId).description
+        let preferenceList = Topic.preferenceList(validId).description
+        let conversations = fixtures.bobClient.conversations
```
nakajima, line 94: If we moved `isValidTopic` to `Topic` then we wouldn't need to bring in fixtures or deal with awaits below.

Tests/XMTPTests/ConversationsTest.swift
```diff
@@ -78,4 +78,64 @@ class ConversationsTests: XCTestCase {
...
+        // check if validation of topics accepts all types
+        Task(priority: .userInitiated) {
+            let resultPrivateStore = await conversations.isValidTopic(topic: privateStore)
+            let resultContact = await conversations.isValidTopic(topic: contact)
+            let resultUserIntro = await conversations.isValidTopic(topic: userIntro)
+            let resultUserInvite = await conversations.isValidTopic(topic: userInvite)
+            let resultDirectMessageV1 = await conversations.isValidTopic(topic: directMessageV1)
+            let resultDirectMessageV2 = await conversations.isValidTopic(topic: directMessageV2)
+            let resultPreferenceList = await conversations.isValidTopic(topic: preferenceList)
+            XCTAssertTrue(
```
nakajima, line 105: Could be nice to have separate assertions in case one of these fails but I don't feel super strong about that.

Sources/XMTP/Messages/Topic.swift
```diff
@@ -39,4 +39,8 @@ public enum Topic {
 	private func wrap(_ value: String) -> String {
 		"/xmtp/0/\(value)/proto"
 	}
+    
+    static func isValidTopic(topic: String) -> Bool {
+        return topic.allSatisfy(\.isASCII)
```
rygine, line 44: also [commented](https://github.com/xmtp/xmtp-android/pull/141#discussion_r1408655561) on the Android PR on this. do we want to allow characters 0-32 in topics?

PR #196, opened 2023-11-27
Add more decrypted message functions

nakajima:
In React Native we need more access to decrypted messages that aren't necessarily decoded. This exposes DecyptedMessage in more places.

### DIFF
--- a/Sources/XMTP/Conversations.swift
+++ b/Sources/XMTP/Conversations.swift
@@ -135,6 +135,46 @@ public actor Conversations {
 		}
 	}
 
+	public func streamAllDecryptedMessages() async throws -> AsyncThrowingStream<DecryptedMessage, Error> {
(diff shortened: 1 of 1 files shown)


Sources/XMTP/Conversations.swift
```diff
@@ -1,321 +1,392 @@
...
+							print("discarding message, unable to decode \(envelope)")
+							return nil
+						}
+					}
+				}
+		}
+		return messages
+	}
+
+	public func listBatchDecryptedMessages(topics: [String: Pagination?]) async throws -> [DecryptedMessage] {
```
nplasterer, line 67: Might be worth writing tests for these new functions.

nplasterer (approved):
You'll probably need to bump the pod file again because it was already released.

PR #199, opened 2023-11-27
Fix example app

nakajima:
I just removed and re-added the XMTP dep inside Xcode and it seemed to work?


nplasterer (approved):

daria-github (approved):
Example app works great now, thanks for making the fix!

PR #200, opened 2023-11-28
make canMessage method accessible on the Client class

kele-leanes:
This PR exposes the `canMessage` method as requested in this [comment](https://github.com/xmtp/xmtp-react-native/issues/109#issuecomment-1719252634)

### DIFF
--- a/Tests/XMTPTests/ClientTests.swift
+++ b/Tests/XMTPTests/ClientTests.swift
@@ -30,20 +30,20 @@ class ClientTests: XCTestCase {
 		XCTAssertFalse(cannotMessage)
 	}
 
-  func testStaticCanMessage() async throws {
-      try TestConfig.skip(because: "run manually against dev")
-    let opts = ClientOptions(api: ClientOptions.Api(env: .local, isSecure: false))
+    func testStaticCanMessage() async throws {
+        try TestConfig.skip(because: "run manually against local")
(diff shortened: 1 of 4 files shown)


Tests/XMTPTests/ClientTests.swift
```diff
@@ -30,6 +30,20 @@ class ClientTests: XCTestCase {
 		XCTAssertFalse(cannotMessage)
 	}
 
+  func testStaticCanMessage() async throws {
+      try TestConfig.skip(because: "run manually against dev")
+    let opts = ClientOptions(api: ClientOptions.Api(env: .local, isSecure: false))
```
nplasterer, line 35: It seems like this is running against local not dev?
```suggestion
      try TestConfig.skip(because: "run manually against local")
    let opts = ClientOptions(api: ClientOptions.Api(env: .local, isSecure: false))
```

kele-leanes: if you run the test locally you will get the following error:

`error: -[XMTPTests.ClientTests testStaticCanMessage] : failed: caught error: "creationError("setup error")"`

you should run it pointing to `dev` to pass the test

nplasterer:
@kele-leanes This is looking good you mind bumping the pod spec like I did here: https://github.com/xmtp/xmtp-ios/pull/193 So that you can pick up these changes in React Native 🙏

nakajima (approved):

PR #204, opened 2023-12-06
Disconnect and reconnect stream on background

fabriguespe:
After running some tests, the `stream` fails silently after 4-10 minutes of the app being in the background.

I've implemented a solution addressing streams intermittently stopping after the app is backgrounded and re-foregrounded. The solution involves listening for specific app state notifications and managing the stream accordingly.

Here's the relevant code:

```swift
// Rest of the code ...
func startStream() {
  streamTask = Task {
	  do {
		  for try await message in conversation.streamMessages() {
			  let content: String = try message.content()
			  print("Received message: \(content)")
			  await MainActor.run {
				  messages.append(message)
			  }
		  }
	  } catch {
		  print("Error in message stream: \(error)")
	  }
     }	
}
// Rest of the code ...
.onReceive(NotificationCenter.default.publisher(for: UIApplication.willResignActiveNotification)) { in
 streamTask?.cancel()
}
.onReceive(NotificationCenter.default.publisher(for: UIApplication.willEnterForegroundNotification)) { in
  Task{
    await loadMessages()
    startStreamTask()
  }
}
```
In this code:

1. We listen for the `UIApplication.willResignActiveNotification` notification. When this notification is received, indicating the app is moving to the background, we cancel the `streamTask`, effectively stopping the stream.

2. We also listen for the `UIApplication.willEnterForegroundNotification` notification. When this notification is received, indicating the app is moving to the foreground, we create a new task that loads the messages and starts the stream again.

This approach ensures that the stream is properly managed when the app transitions between the foreground and background states, preventing it from 'dying' silently without errors.


fabriguespe:
Added a method for starting the app from a private key for faster testing

```swift
    func dataFromHexString(_ hex: String) -> Data? {
        var data = Data(capacity: hex.count / 2)
        var buffer = 0
        var index = 0
        for char in hex {
            if let value = char.hexDigitValue {
                if index % 2 == 0 {
                    buffer = value << 4
                } else {
                    buffer |= value
                    data.append(UInt8(buffer))
                }
                index += 1
            } else {
                return nil
            }
        }
        return data
    }
	func generateWalletWithPrivateKeys() {
    Task {
        do {
			let privateKeyString = "key" // Your PrivateKey instance here
            // Function to convert hex string to Data
            
            
            if let privateKeyData = dataFromHexString(privateKeyString) {
				let privateKey = try PrivateKey(privateKeyData)
            let client = try await Client.create(account: privateKey, options: ClientOptions(api: .init(env: .production)))
            
			
            print(client.address)
            let keysData = try client.privateKeyBundle.serializedData()
            Persistence().saveKeys(keysData)

            await MainActor.run {
                self.status = .connected(client)
            }
			} else {
				// Handle the error case where the hex string couldn't be converted to Data
			}
           
            
        } catch {
            await MainActor.run {
                self.status = .error("Error generating wallet: \(error)")
            }
        }
    }
	}
```

XMTPiOSExample/XMTPiOSExample/ContentView.swift
```diff
@@ -66,12 +53,113 @@ struct ContentView: View {
...
+    }
+	}
 	func generateWallet() {
 		Task {
 			do {
 				let wallet = try PrivateKey.generate()
-				let client = try await Client.create(account: wallet, options: .init(api: .init(env: .dev, isSecure: true, appVersion: "XMTPTest/v1.0.0")))
-
+				let client = try await Client.create(account: wallet, options: .init(api: .init(env: .production, isSecure: true, appVersion: "XMTPTest/v1.0.0")))
+                print (client.address);
```
nakajima, line 162: Do we need these changes?

XMTPiOSExample/XMTPiOSExample/ContentView.swift
```diff
@@ -66,12 +53,113 @@ struct ContentView: View {
...
+            } else {
+                return nil
+            }
+        }
+        return data
+    }
+	func generateWalletWithPrivateKeys() {
+    Task {
+        do {
+			let privateKeyString = "67633be8c32db5414951db4a9ea9734b1214f8f5ca15d6b16818c0b4ee864653" // Your PrivateKey instance here
```
nakajima, line 129: Do we actually want to commit this private key?

PR #207, opened 2023-12-07
Generate docs content to Pages

fabriguespe:
Fixed SDK docs [issues](https://github.com/xmtp/xmtp-ios/issues) while doing docs deployment

- [Preview](https://xmtp.github.io/xmtp-ios/)

```yml
name: Deploy docs content to GitHub Pages

on:
  push:
    branches: ["main", "docs", "docs_build_v1"]
  workflow_dispatch:

permissions:
  contents: read
  pages: write
  id-token: write

concurrency:
  group: "pages"
  cancel-in-progress: false

jobs:
  deploy:
    environment:
      name: github-pages
      url: ${{ steps.deployment.outputs.page_url }}
    runs-on: macos-12
    steps:
      - uses: actions/checkout@v3

      - name: Set up Ruby
        uses: ruby/setup-ruby@v1
        with:
          ruby-version: 3.0

      - name: Cache Ruby gems
        uses: actions/cache@v2
        with:
          path: vendor/bundle
          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
          restore-keys: |
            ${{ runner.os }}-gems-

      - name: Install Jazzy
        run: |
          bundle config path vendor/bundle
          bundle add jazzy

      - name: Generate documentation
        run: bundle exec jazzy --output ./docs --theme=fullwidth

      - name: Setup Pages
        uses: actions/configure-pages@v3

      - name: Upload artifact
        uses: actions/upload-pages-artifact@v1
        with:
          path: "./docs"

      - name: Deploy to GitHub Pages
        id: deployment
        uses: actions/deploy-pages@v2
```

To run this locally:

```
gem install jazzy
jazzy --output ./docs --theme=fullwidth
```


.github/workflows/docs.yml
```diff
@@ -0,0 +1,57 @@
+name: Deploy docs content to Pages 2
+
+on:
+  push:
+    branches: ["docs_build_v1"]
```
nplasterer, line 5: We probably want this to actually normally happen on pushes to main 

```suggestion
    branches: ["main", "docs_build_v1"]
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,57 @@
+name: Deploy docs content to Pages 2
```
nplasterer, line 1:
```suggestion
name: Deploy docs content to GitHub Pages
```

Sources/XMTP/Client.swift
```diff
@@ -42,7 +42,7 @@ public struct ClientOptions {
 	}
 }
 
-/// Client is the entrypoint into the XMTP SDK.
+/// Client is the entrypoint into the XMTP SDK.  v2
```
nplasterer, line 45:
```suggestion
/// Client is the entrypoint into the XMTP SDK.
```

Gemfile
```diff
@@ -0,0 +1,2 @@
+source 'https://rubygems.org'
+gem 'jazzy'
```
nplasterer, line 2: I don't know enough about iOS development to know if this would be an issue cc @nakajima

nplasterer (commented):
Nice! Seems to be working. I'd suggest deleting the old docs deployment action so we don't have two running https://github.com/xmtp/xmtp-ios/blob/main/.github/workflows/build-docs.yml

Gemfile
nakajima: Seems fine to me.

nakajima (approved):

PR #212, opened 2024-01-02
Custom Content Codecs when preparing messages

alexrisch:
Adds support for prepareMessage to handled encoded content

### DIFF
--- a/Sources/XMTP/Conversation.swift
+++ b/Sources/XMTP/Conversation.swift
@@ -136,6 +136,15 @@ public enum Conversation: Sendable {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Sources/XMTP/Conversation.swift
```diff
@@ -136,6 +136,15 @@ public enum Conversation: Sendable {
 		}
 	}
 
+  @discardableResult public func prepareMessage(encodedContent: EncodedContent, options: SendOptions? = nil) async throws -> PreparedMessage {
```
nplasterer, line 139: Is there a reason to make the result discardable here?

nplasterer (approved):
Don't forget you'll need to bump the podspec so that we can consume this in RN when it merges.

PR #214, opened 2024-01-16
fix: Improve performance list method

kele-leanes:
**Description:**
Addressing the concerns raised in [issue #130](https://github.com/xmtp/xmtp-react-native/issues/130), this pull request enhances the pagination functionality for the list method. The key improvement involves introducing a new optional parameter, `createdAt`, which can be set when a conversation is created. This parameter is subsequently utilized in the `importToTopic` method to establish a reference point, enabling the retrieval of the most recent conversations from that specific date. Incorporating this enhancement significantly contributes to optimizing the request time for improved overall performance. 

| 0 conversations imported | 500 conversations imported|
|:--:|:--:|
| Screenshot 2024-01-16 at 16 11 10 | Screenshot 2024-01-16 at 16 13 46 |

### DIFF
--- a/Sources/XMTP/ConversationV2.swift
+++ b/Sources/XMTP/ConversationV2.swift
@@ -15,11 +15,12 @@ public struct ConversationV2Container: Codable {
 	var conversationID: String?
 	var metadata: [String: String] = [:]
 	var peerAddress: String
+	var createdAt: UInt64?
 	var header: SealedInvitationHeaderV1
 
 	public func decode(with client: Client) -> ConversationV2 {
 		let context = InvitationV1.Context(conversationID: conversationID ?? "", metadata: metadata)
-		return ConversationV2(topic: topic, keyMaterial: keyMaterial, context: context, peerAddress: peerAddress, client: client, header: header)
+		return ConversationV2(topic: topic, keyMaterial: keyMaterial, context: context, peerAddress: peerAddress, client: client, header: header, createdAt: createdAt ?? 0)
 	}
 }
 
@@ -30,6 +31,7 @@ public struct ConversationV2 {
 	public var context: InvitationV1.Context
 	public var peerAddress: String
 	public var client: Client
+	public var createdAt: UInt64?
 	private var header: SealedInvitationHeaderV1
 
 	static func create(client: Client, invitation: InvitationV1, header: SealedInvitationHeaderV1) throws -> ConversationV2 {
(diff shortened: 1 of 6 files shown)


Sources/XMTP/ConversationV2.swift
```diff
@@ -188,8 +193,8 @@ public struct ConversationV2 {
 		}
 	}
 
-	public var createdAt: Date {
-		Date(timeIntervalSince1970: Double(header.createdNs / 1_000_000) / 1000)
+	public var conversationCreatedAt: Date {
+		Date(timeIntervalSince1970: Double((createdAt ?? 0) / 1_000_000) / 1000)
```
nplasterer, line 197: I think if you don't have the parent createdAt then you should use the header
```suggestion
		Date(timeIntervalSince1970: Double((createdAt ?? header.createdNs) / 1_000_000) / 1000)
```

Sources/XMTP/ConversationV2.swift
```diff
@@ -15,11 +15,12 @@ public struct ConversationV2Container: Codable {
 	var conversationID: String?
 	var metadata: [String: String] = [:]
 	var peerAddress: String
+	var createdAt: UInt64?
 	var header: SealedInvitationHeaderV1
 
 	public func decode(with client: Client) -> ConversationV2 {
 		let context = InvitationV1.Context(conversationID: conversationID ?? "", metadata: metadata)
-		return ConversationV2(topic: topic, keyMaterial: keyMaterial, context: context, peerAddress: peerAddress, client: client, header: header)
+		return ConversationV2(topic: topic, keyMaterial: keyMaterial, context: context, peerAddress: peerAddress, client: client, createdAt: createdAt ?? 0, header: header)
```
nplasterer, line 23: This allows createdAt to be nil so let it be nil.

```suggestion
		return ConversationV2(topic: topic, keyMaterial: keyMaterial, context: context, peerAddress: peerAddress, client: client, createdAt: createdAt, header: header)
```

Sources/XMTP/ConversationV2.swift
```diff
@@ -188,8 +193,8 @@ public struct ConversationV2 {
 		}
 	}
 
-	public var createdAt: Date {
-		Date(timeIntervalSince1970: Double(header.createdNs / 1_000_000) / 1000)
+	public var conversationCreatedAt: Date {
```
nplasterer, line 196: Do we still need this? Maybe we can move this logic up? Otherwise I'd say lets not rename this here. I think if this is whats being accessed externally it should stay createdAt
```suggestion
	public var createdAt: Date {
```

Sources/XMTP/ConversationV2.swift
```diff
@@ -30,6 +31,7 @@ public struct ConversationV2 {
 	public var context: InvitationV1.Context
 	public var peerAddress: String
 	public var client: Client
+	public var createdAt: UInt64?
```
nplasterer, line 34: If this doesn't get accessed externally lets rename this one instead. and make it private. This is just the nanoseconds for when it was created so that may be more descriptive for others using it.
```suggestion
	var createdAtNs: UInt64?
```

Tests/XMTPTests/ConversationsTest.swift
```diff
@@ -122,4 +122,39 @@ class ConversationsTests: XCTestCase {
         XCTAssertFalse(Topic.isValidTopic(topic: directMessageV2))
         XCTAssertFalse(Topic.isValidTopic(topic: preferenceList))
     }
+	
+	func testLoadConvos() async throws {
```
nplasterer, line 126: I wouldn't commit this test just because it uses a hardcoded wallet on dev that could change at any moment. But appreciate you thoroughly testing it.

nplasterer (commented):
Just a few comments. Also don't forget to bump the podSpec so we can use this in React Native.

Sources/XMTP/ConversationV2.swift
kele-leanes: What are you referring to moving up? This method is needed to get the most recent saved conversation. I would rename it as you suggested

kele-leanes:
> Just a few comments. Also don't forget to bump the podSpec so we can use this in React Native.

Thanks for your review. Let me know if there is something else

Sources/XMTP/ConversationV2.swift
```diff
@@ -15,11 +15,12 @@ public struct ConversationV2Container: Codable {
 	var conversationID: String?
 	var metadata: [String: String] = [:]
 	var peerAddress: String
+	var createdAt: UInt64?
```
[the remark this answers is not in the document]
kele-leanes: Good catch!

nplasterer (approved):
LGTM. 👍

PR #217, opened 2024-01-17
Replace xmtp-rust-swift with libxmtp-swift

nakajima:
This PR also renames the XMTP module to XMTPiOS.

Relies on https://github.com/xmtp/libxmtp-swift/pull/2.


XMTP.podspec
```diff
@@ -44,5 +44,5 @@ Pod::Spec.new do |spec|
   spec.dependency "web3.swift"
   spec.dependency "GzipSwift"
   spec.dependency "Connect-Swift", "= 0.3.0"
-  spec.dependency 'XMTPRust', '= 0.3.7-beta0'
+  spec.dependency 'LibXMTP., '= 0.3.7-beta0'
```
nplasterer, line 47: Is the period necessary here?

Sources/XMTPiOS/ApiClient.swift
```diff
@@ -0,0 +1,174 @@
...
+					//                    // Run a continuous for loop polling and sleeping for a bit each loop.
+					//                    while true {
+					//                        let buf = try subscription.get_envelopes_as_query_response()
+					//                        // Note: it uses QueryResponse as a convenient envelopes wrapper.
+					//                        let res = try QueryResponse(serializedData: Data(buf))
+					//                        for envelope in res.envelopes {
+					//                            continuation.yield(envelope)
+					//                        }
+					//                        try await Task.sleep(nanoseconds: 50_000_000) // 50ms
+					//                    }
```
nplasterer, line 153: Was this not implemented before? Just want to make sure we're not breaking subscribe.

nplasterer (approved):
Seems like this will be a breaking change for any clients already integrated just a rename from `XMTP` to `XMTPiOS`. Should make sure to write up the change in the release notes for anyone who has issues.

XMTP.podspec
nplasterer, line 47:
```suggestion
  spec.dependency 'LibXMTP', '= 0.3.7-beta0'
```

Sources/XMTPiOS/ApiClient.swift
nakajima: Oh it was. I should get this working before merging.

neekolas: I can help with the Rust side if needed

XMTP.podspec
```diff
@@ -44,5 +44,5 @@ Pod::Spec.new do |spec|
   spec.dependency "web3.swift"
   spec.dependency "GzipSwift"
   spec.dependency "Connect-Swift", "= 0.3.0"
-  spec.dependency 'XMTPRust', '= 0.3.7-beta0'
+  spec.dependency 'LibXMTP', '= 0.3.7-beta0'
```
cameronvoell, line 47:
```suggestion
  spec.dependency 'LibXMTP', '= 0.0.1-timeout0'
```

Set to the verision from [LibXMTP podspec](https://github.com/xmtp/libxmtp-swift/blob/main/LibXMTP.podspec#L3)

nplasterer: Thats an odd name. 😄 probably something we want to fix eventually.

PR #218, opened 2024-01-19
Add sender HMAC to MessageV2

nplasterer:
Closes https://github.com/xmtp/xmtp-ios/issues/216

Follows the pattern put in place here: https://github.com/xmtp/xmtp-js/pull/519

This fixes the iOS entitlement issue for push notifications by adding additional information to the payload. See: https://github.com/xmtp/XIPs/pull/35


Sources/XMTP/Codecs/ReplyCodec.swift
```diff
@@ -66,4 +66,8 @@ public struct ReplyCodec: ContentCodec {
     public func fallback(content: Reply) throws -> String? {
         return "Replied with “\(content.content)” to an earlier message"
     }
+	
+	public func shouldPush(content: Reply) throws -> Bool {
+		return false
```
nakajima, line 71: Do we not want to push for replies?

Sources/XMTP/Codecs/RemoteAttachmentCodec.swift
```diff
@@ -206,4 +206,8 @@ public struct RemoteAttachmentCodec: ContentCodec {
 
 		return Data(parameterData)
 	}
+	
+	public func shouldPush(content: RemoteAttachment) throws -> Bool {
+		return false
```
nakajima, line 211: Do we not want to push for remote attachments?

Tests/XMTPTests/MessageTests.swift
```diff
@@ -44,7 +44,7 @@ class MessageTests: XCTestCase {
 	}
 
 	func testFullyEncodesDecodesMessagesV2() async throws {
-	    try TestConfig.skipIfNotRunningLocalNodeTests()
+//	    try TestConfig.skipIfNotRunningLocalNodeTests()
```
nakajima, line 47: Should we bring this back?

Sources/XMTP/Codecs/ReplyCodec.swift
kele-leanes: the scope of the task only includes `textCodec` and `Composite`

Sources/XMTP/Codecs/RemoteAttachmentCodec.swift
kele-leanes: the scope of the task only includes `textCodec` and `Composite`

nplasterer: 👋 @kele-leanes I actually shared a table with you that showed the rest of the codecs as well. We definitely want this to be true. Just because JS has codecs in a different repo doesnt mean they aren't also supported. Here is the table for reference again.

Screenshot 2024-01-30 at 7 26 54 AM

Tests/XMTPTests/CodecTests.swift
```diff
@@ -9,6 +9,10 @@ import XCTest
 @testable import XMTPiOS
 
 struct NumberCodec: ContentCodec {
+	func shouldPush(content: Double) throws -> Bool {
```
nplasterer, line 12: Would you mind adding a test in here that shows you can get the shouldPush method correctly from the decodedMessage kind of like we should have fallback tests. 🙏

nplasterer (commented):
Looks like you have two legit linter issues

`/Users/runner/work/xmtp-ios/xmtp-ios/Sources/XMTPiOS/Messages/MessageV2.swift:105:91: error: Force Unwrapping Violation: Force unwrapping should be avoided `

`/Users/runner/work/xmtp-ios/xmtp-ios/Sources/XMTPiOS/ConversationV2.swift:249:50: error: Force Cast Violation: Force casts should be avoided (force_cast)`

nplasterer:
I can't approve this since I opened it but would be good to get eyes from @rygine and or @Bren2010 on the cryptography parts and @nakajima on the iOS code 🙏  But looks good to me.

rygine:
the crypto code _seems_ OK, but i'll defer to someone with more knowledge. all the other pieces look to be in the right places.

Tests/XMTPTests/CodecTests.swift
kele-leanes: seems that `Xmtp_MessageContents_EncodedContent` does not have the `shouldPush` method available which is where the fallback is coming from. Should I get it from there or from a different place?

nplasterer: That was my bad from looking at JS it's on the messagev2 so wrote a test for that scenario and pushed it up. https://github.com/xmtp/xmtp-ios/pull/218/commits/1967934ae85789b62885e2850de6e008840eb97b

nakajima (approved):

nplasterer:
@kele-leanes do you mind bumping the pod spec so we can consume this in RN?

Tests/XMTPTests/CodecTests.swift
kele-leanes: thanks! @nplasterer

PR #226, opened 2024-01-29
Subscribe2

nplasterer:
WIP
Improvements to subscribeAllMessages use the subscription to update the stream instead of breaking and resubscribing.

All the tests pass but I think we just don't have great testing around this.


Sources/XMTPiOS/Conversations.swift
```diff
@@ -99,33 +99,33 @@ public actor Conversations {
...
+				]
 
-					for conversation in try await list() {
-						topics.append(conversation.topic)
-					}
+				for conversation in try await list() {
+					topics.append(conversation.topic)
+				}
 
+				while(true) {
```
nplasterer, line 111: I'm not sure where this while loop should actually live. 🤔

Sources/XMTPiOS/Conversations.swift
```diff
@@ -99,33 +99,33 @@ public actor Conversations {
...
+						for try await (envelope, subscription) in client.subscribe(topics: topics) {
 							if let conversation = conversationsByTopic[envelope.contentTopic] {
 								let decoded = try conversation.decode(envelope)
 								continuation.yield(decoded)
 							} else if envelope.contentTopic.hasPrefix("/xmtp/0/invite-") {
 								let conversation = try fromInvite(envelope: envelope)
 								conversationsByTopic[conversation.topic] = conversation
-								break // Break so we can resubscribe with the new conversation
+								topics.append(conversation.topic)
+								try await subscription.update(req: client.makeSubscribeRequest(topics: topics).toFFI)
```
nplasterer, line 121: Does this update method want the entire list of topics or just the new topics?

Sources/XMTPiOS/ApiClient.swift
```diff
@@ -132,8 +133,12 @@ final class GRPCApiClient: ApiClient {
 
 		return envelopes
 	}
+	
+	func makeSubscribeRequest(topics: [String]) -> SubscribeRequest {
+		return SubscribeRequest.with { $0.contentTopics = topics }
+	}
 
-	func subscribe(topics: [String]) -> AsyncThrowingStream<Envelope, Error> {
+	func subscribe(topics: [String]) -> AsyncThrowingStream<(envelope: Envelope, subscription: LibXMTP.FfiV2Subscription), Error> {
```
nplasterer, line 141: A Touple probably isn't the best idea to solve this. But the simplest for now.

Sources/XMTPiOS/Conversations.swift
```diff
@@ -99,33 +99,33 @@ public actor Conversations {
...
-						topics.append(conversation.topic)
-					}
+				for conversation in try await list() {
+					topics.append(conversation.topic)
+				}
 
+				while(true) {
 					do {
-						for try await envelope in client.subscribe(topics: topics) {
+						for try await (envelope, subscription) in client.subscribe(topics: topics) {
```
nplasterer, line 113: I tried a scenario where we returned the subscription independent of this for loop but the await function never completed. It may be my lack of knowledge of iOS threads/streams. But you can see that work in the previous commit history.

nplasterer:
@fabriguespe the tests definitely pass but I'm suspicious. Do you have an example you're running to test this? Can we try it on this branch?

Sources/XMTPiOS/Conversations.swift
nplasterer: This method is also blocking from the for loop continuing..

nplasterer:
Okay I was able to reproduce that this isn't fixing the stream entirely. I've got an issue with subcription.update blocking the thread from continueing and when I put it in a task it doesn't actually update. You can see the failure in this test

https://github.com/xmtp/xmtp-ios/blob/main/Tests/XMTPTests/IntegrationTests.swift#L248-L277

nplasterer:
This may be a red herring. If everyone thinks this improvement is futile I'm happy to abandon this effort.

Sources/XMTPiOS/Conversations.swift
neekolas: It takes the entire list of topics

neekolas (approved):
Approved for whenever you think this is ready to go

Sources/XMTPiOS/Conversations.swift
nplasterer: @zombieobject I've got some iOS threading questions for you next week 🙏

nplasterer:
@zombieobject and I paired on this today and decided that this probably isn't worth our time right now to fix. I think the real fix will likely be in the rust layer changing the current streaming interface to a callback interface like the V3 work does. Since this works fine for now and we have other high priorities going to pause this work.

PR #229, opened 2024-02-01
Introduce group chat to xmtp-ios

nakajima:
Begins to introduce group chats to xmtp-ios. Right now, chats are loaded by calling `client.conversations.groups()`. The example app has been updated to call this and combine groups with conversations under the `ConversationOrGroup` enum. Once this PR lands, we'll embark on a broader change where `Conversation` is renamed to `DirectMessage` which will free up the `Conversation` name to be a union of `DirectMessage` and `Group`.

TODO:

- ~~The most recent version of libxmtp-swift bindings didn't build correctly for me so I [switched back to building a dylib in my local libxmtp](https://gist.github.com/nakajima/08337d23f8de5e3feeea64bb80a0a382) for now. It was missing the [LegacyIdentitySource](https://github.com/xmtp/libxmtp/blob/main/bindings_ffi/src/mls.rs#L24-L39) stuff needed for client creation. We'll need to work that out before merging.~~ done
- ~~Message/Group Pagination~~ done
- ~~Handle GroupMemberChanged messages in the example app~~ done
- Group Streaming (could happen in a follow up)

Here's some screenshots from the example app:

|Creating a group|Viewing a group|Messages|Group settings|
|-|-|-|-|
|Screenshot 2024-02-01 at 3 06 57 PM|Screenshot 2024-02-01 at 3 06 57 PM|Simulator Screenshot - iPhone 15 Plus - 2024-02-06 at 13 12 08|image|


Sources/XMTPiOS/CodecRegistry.swift
```diff
@@ -8,7 +8,10 @@
 import Foundation
 
 struct CodecRegistry {
-	var codecs: [String: any ContentCodec] = [TextCodec().id: TextCodec()]
+	var codecs: [String: any ContentCodec] = [
+		TextCodec().id: TextCodec(),
+		GroupMembershipChangedCodec().id: GroupMembershipChangedCodec()
```
nplasterer, line 13: @richardhuaaa and I discussed this yesterday and we actually think it doesn't make sense to auto register `GroupMembershipChangedCodec` since it would require rendering code to be supported on the client side and would break if they don't support that.

Sources/XMTPiOS/Conversations.swift
```diff
@@ -13,6 +17,20 @@ public actor Conversations {
 		self.client = client
 	}
 
+	public func groups() async throws -> [Group] {
+		try await client.v3Client.conversations().sync()
```
nplasterer, line 22: We decided to not auto sync groups on groups load and instead groups should take pagination data.

Sources/XMTPiOS/Conversations.swift
```diff
@@ -13,6 +17,20 @@ public actor Conversations {
...
+
+		return try await client.v3Client.conversations().list(opts: .init(createdAfterNs: nil, createdBeforeNs: nil, limit: nil)).map { $0.fromFFI(client: client) }
+	}
+
+	public func newGroup(with addresses: [String]) async throws -> Group {
+		if addresses.isEmpty {
+			throw GroupError.emptyCreation
+		}
+
```
nplasterer, line 45: I did a bunch of checks here in android that would probably be worth doing here as well

https://github.com/xmtp/xmtp-android/pull/160/files#diff-479fe53ececf1836cd910bfba9a721b20fce9fd212b204329abef95ff79429f9R90-R105

Sources/XMTPiOS/Group.swift
```diff
@@ -0,0 +1,106 @@
...
+		return try ffiGroup.listMembers().map(\.fromFFI)
+	}
+
+	public var cachedMembers: [Member] {
+		do {
+			return try ffiGroup.listMembers().map(\.fromFFI)
+		} catch {
+			return []
+		}
+	}
```
nplasterer, line 46: I ended up just make a list of member addresses instead of a list of members what else is on the member is useful to integrators?

Sources/XMTPiOS/Group.swift
```diff
@@ -0,0 +1,106 @@
...
+
+	public func addMembers(addresses: [String]) async throws {
+		try await ffiGroup.addMembers(accountAddresses: addresses)
+		try await ffiGroup.sync()
```
nplasterer, line 50: I think calling sync in all these places is probably hurting the performance? I just expose the sync method instead.

Sources/XMTPiOS/Group.swift
```diff
@@ -0,0 +1,106 @@
...
+		if let compression = options?.compression {
+			encoded = try encoded.compress(compression)
+		}
+
+		try await ffiGroup.send(contentBytes: encoded.serializedData())
+	}
+
+	public func messages() async throws -> [DecodedMessage] {
+		// TODO: paginate
+		try await ffiGroup.sync()
```
nplasterer, line 90: I think we should opt for pagination instead of adding sync here.

Sources/XMTPiOS/XMTPLogger.swift
```diff
@@ -0,0 +1,18 @@
...
+
+import Foundation
+import LibXMTP
+import os
+
+class XMTPLogger: FfiLogger {
+	let logger = Logger()
+
+	func log(level: UInt32, levelLabel: String, message: String) {
+		logger.info("libxmtp[\(levelLabel)] - \(message)")
```
nplasterer, line 16: It would be nice to not always log on info and instead switch on level so
if level is 3 then logger.info

This is how Rust maps

```
Error: 1
Warn: 2
Info: 3
Debug: 4
Trace: 5
```

Sources/XMTPiOS/Client.swift
```diff
@@ -155,6 +173,10 @@ public final class Client: Sendable {
 		return nil
 	}
 
+	public func canMessageV3(address: String) async throws -> Bool {
+		try await v3Client.canMessage(accountAddresses: [address]) == [true]
+	}
+
 	public static func from(bundle: PrivateKeyBundle, options: ClientOptions? = nil) async throws -> Client {
```
nplasterer, line 180: It's hard to see here but did you update the from bundle method to support V3 as well?

Sources/XMTPiOS/Client.swift
```diff
@@ -100,23 +101,40 @@ public final class Client: Sendable {
 	}
 
 	static func create(account: SigningKey, apiClient: ApiClient, options: ClientOptions? = nil) async throws -> Client {
-		let privateKeyBundleV1 = try await loadOrCreateKeys(for: account, apiClient: apiClient, options: options)
+		let (privateKeyBundleV1, source) = try await loadOrCreateKeys(for: account, apiClient: apiClient, options: options)
+
+		let dbURL = URL.documentsDirectory.appendingPathComponent("xmtp-\(account.address).db3")
+		let v3Client = try await LibXMTP.createClient(
```
nplasterer, line 107: V3 creation should not happen unless a flag is set in the ClientOptions and the environment is Local. Be good to safe guard that.

Sources/XMTPiOS/Client.swift
```diff
@@ -100,23 +101,40 @@ public final class Client: Sendable {
...
+			logger: XMTPLogger(),
+			host: GRPCApiClient.envToUrl(env: apiClient.environment),
+			isSecure: apiClient.environment != .local,
+			db: dbURL.path,
+			encryptionKey: nil,
+			accountAddress: account.address,
+			legacyIdentitySource: source,
+			legacySignedPrivateKeyProto: nil
```
nplasterer, line 115: I don't think this should be nil?

Sources/XMTPiOS/CodecRegistry.swift
neekolas: That makes sense to me.

We should discuss if we want fallback text here. I'm leaning towards no, since it's a little weird to just have auto-generated messages that you didn't send. If these aren't rendered distinctly from normal conversation messages I think the experience is going to be weird.

nakajima: Yea, when I do the rename of Conversation -> DirectMessage then I'll introduce a flag to officially enable group chat and we can conditionally register this codec then.

For now, the only time a user would ever see this type of content is if they explicitly load groups via `conversations.groups()`, the old `list` method is unchanged here.

richardhuaaa: > we can conditionally register this codec then

So it might make sense to actually let the app developer register this codec, if and only if they add code to handle the codec - even if they enable group chats their app might crash otherwise.

Sources/XMTPiOS/Client.swift
```diff
@@ -100,23 +101,40 @@ public final class Client: Sendable {
...
+		)
+
+		guard let textToSign = v3Client.textToSign() else {
+			throw ClientError.creationError("no text to sign")
```
nplasterer, line 119: It's a totally okay scenario to not have text to sign. That actually means you already have V3 keys. We shouldn't error here.

cameronvoell:
> The most recent version of libxmtp-swift bindings didn't build correctly for me so I [switched back to building a dylib in my local libxmtp](https://gist.github.com/nakajima/08337d23f8de5e3feeea64bb80a0a382) for now. It was missing the [LegacyIdentitySource](https://github.com/xmtp/libxmtp/blob/main/bindings_ffi/src/mls.rs#L24-L39) stuff needed for client creation. We'll need to work that out before merging.

@nakajima If you let me know what commit in [libxmtp](https://github.com/xmtp/libxmtp) you would like bindings from, I should be able to create static bindings that will work for you in Package.swift. Let me know and I can add it as a PR on to this one.

nakajima:
@cameronvoell I _think_ whatever is on main should have everything I need? So https://github.com/xmtp/libxmtp/commit/60ee7556e8c49ec0bf3e2e23c80e8fe4006b11a0 should do it.

Sources/XMTPiOS/Group.swift
nakajima: For now I think member addresses is the only interesting thing, tho I could see that changing if we ever introduce some notion of nicknames or maybe who added whom. I was mostly just mirror LibXMTP here but if we just want addresses for consistency between ios and android I can make the change.

Sources/XMTPiOS/Client.swift
nakajima: Ya, it's updated in the `from` that this one calls.

Sources/XMTPiOS/Client.swift
nakajima: yea I misunderstood what it's from. Updated to pass that in.

Sources/XMTPiOS/Client.swift
nakajima: Good to know, updated.

Sources/XMTPiOS/Client.swift
nakajima: Added in 04a95aa83977578b32aa14030fed3518748fb83f.

Sources/XMTPiOS/CodecRegistry.swift
nakajima: Ok I'm sold, making it opt-in.

Sources/XMTPiOS/Conversations.swift
nakajima: I was relying on the errors bubbling up from libxmtp but I think you're right, having our own errors is nicer. Updated

Sources/XMTPiOS/XMTPLogger.swift
nakajima: Updated

nakajima:
@cameronvoell Things seem to work pointing at https://github.com/xmtp/libxmtp-swift/pull/6. Thanks for the libxmtp fix!

Package.swift
```diff
@@ -20,14 +20,12 @@ let package = Package(
...
 		// .package(url: /* package url */, from: "1.0.0"),
-
 		.package(url: "https://github.com/GigaBitcoin/secp256k1.swift.git", exact: "0.10.0"),
 		.package(url: "https://github.com/argentlabs/web3.swift", from: "1.1.0"),
 		.package(url: "https://github.com/1024jp/GzipSwift", from: "5.2.0"),
 		.package(url: "https://github.com/bufbuild/connect-swift", exact: "0.3.0"),
 		.package(url: "https://github.com/apple/swift-docc-plugin.git", from: "1.0.0"),
-		.package(url: "https://github.com/xmtp/libxmtp-swift", revision: "92274fe"),
-//		.package(path: "../libxmtp-swift")
+		.package(url: "https://github.com/xmtp/libxmtp-swift", revision: "1d7068f181a402b2cd8b9520ccb768e65a55ec32"),
```
cameronvoell, line 28: I tagged it on main branch at [503086d](https://github.com/xmtp/libxmtp-swift/commit/503086d91ba6c3420aca118e4812d62a004d17ee), after merging your PR, so you'll prob want to update to that commit
```suggestion
		.package(url: "https://github.com/xmtp/libxmtp-swift", revision: "503086d"),
```

cameronvoell:
Tested on RN via PR here https://github.com/xmtp/xmtp-react-native/pull/234/files. Standard V2 features all look good. Looking forward to integrating ios groups on RN, nice work!

Sources/XMTPiOS/CodecRegistry.swift
nplasterer: Thats a valid point but I still don't think we should auto opt-in I think instead maybe we need a fallback text.

Sources/XMTPiOS/Group.swift
nplasterer: Since memberAdd and memberRemove take an address and not a member object. I think right now it makes more sense to just have a list of addresses. In the future if members become more advanced than we can handle that with a new class.

Sources/XMTPiOS/Client.swift
```diff
@@ -100,23 +110,48 @@ public final class Client: Sendable {
 	}
 
 	static func create(account: SigningKey, apiClient: ApiClient, options: ClientOptions? = nil) async throws -> Client {
-		let privateKeyBundleV1 = try await loadOrCreateKeys(for: account, apiClient: apiClient, options: options)
+		let (privateKeyBundleV1, source) = try await loadOrCreateKeys(for: account, apiClient: apiClient, options: options)
+
+		let v3Client: FfiXmtpClient?
+
+		if options?.enableAlphaMLS == true && options?.api.env == .local {
+			let dbURL = URL.documentsDirectory.appendingPathComponent("xmtp-\(account.address).db3")
```
nplasterer, line 118: Nick suggested in my PR also adding the environment so we don't have database conflicts between envs

`val alias = "xmtp-${options!!.api.env}-${account.address.lowercase()}"`

Sources/XMTPiOS/Client.swift
```diff
@@ -100,23 +110,48 @@ public final class Client: Sendable {
...
+			v3Client = try await LibXMTP.createClient(
+				logger: XMTPLogger(),
+				host: GRPCApiClient.envToUrl(env: apiClient.environment),
+				isSecure: apiClient.environment != .local,
+				db: dbURL.path,
+				encryptionKey: nil,
```
nplasterer, line 124: I think it's important to encrypt the database. You can see how I did it here: https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/Client.kt#L300-L322

Sources/XMTPiOS/Client.swift
```diff
@@ -172,13 +215,40 @@ public final class Client: Sendable {
...
+				accountAddress: address,
+				legacyIdentitySource: .static,
+				legacySignedPrivateKeyProto: try v1Bundle.toV2().identityKey.serializedData()
+			)
+
+			try await v3Client?.registerIdentity(recoverableWalletSignature: nil)
```
nplasterer, line 233: This is repeated it would probably be better to just do this in one place so that its harder to introduce bugs. Do you mind pulling this into it's own method?

Sources/XMTPiOS/Client.swift
```diff
@@ -172,13 +215,40 @@ public final class Client: Sendable {
...
+		} else {
+			v3Client = nil
+		}
+
```
nplasterer, line 237: This scenario is actually the one where text to sign is probably going to be most important but I don't see the signer work here?

nplasterer (commented):
Might be nice to pull out the client creation from the rest of the group stuff so that it's easier to review. I just took another pass on that but not the rest of the code.

Sources/XMTPiOS/Client.swift
nakajima: Good call, updating

Sources/XMTPiOS/Client.swift
nakajima: What do you think of allowing a user to pass it in instead of generating it for them?

Sources/XMTPiOS/Client.swift
nakajima: Done

Sources/XMTPiOS/Client.swift
nakajima: Maybe I misunderstood the text to sign parts. This is the flow where we just load keys we already have so we don't have access to a signer?

Sources/XMTPiOS/Client.swift
nplasterer: It will fail if the encryption key isn't the same every time they login or build with bundle. Feels like we should store it for them.

Sources/XMTPiOS/Client.swift
nplasterer: Ya thats the breaking change we discussed last week. If they don't have a signer and also dont have V3 keys they can't create a v3 client. The first time requires a signer. So you'll need to have them pass a signer in which would be a forced log out.

nplasterer: The only way to tell if they have v3 keys or not is to check if they need to sign or not.

Sources/XMTPiOS/Client.swift
nakajima: I feel like persistence of secrets is more of an app level concern, I don't want to be prescriptive about how folks use the keychain since they might also want to load keys from elsewhere.

nakajima:
Pulled the Client changes into their own PR here: https://github.com/xmtp/xmtp-ios/pull/233

Sources/XMTPTestHelpers/TestHelpers.swift
```diff
@@ -23,7 +23,7 @@ public struct TestConfig {
     }
 
     static public func skipIfNotRunningLocalNodeTests() throws {
-        try XCTSkipIf(!TEST_SERVER_ENABLED, "requires local node")
+//        try XCTSkipIf(!TEST_SERVER_ENABLED, "requires local node")
```
nplasterer, line 26: Probably uncomment this

Sources/XMTPiOS/Conversations.swift
```diff
@@ -1,9 +1,14 @@
 import Foundation
+import LibXMTP
 
 public enum ConversationError: Error {
 	case recipientNotOnNetwork, recipientIsSender, v1NotSupported(String)
 }
 
+public enum GroupError: Error {
+	case alphaMLSNotEnabled, emptyCreation, memberCannotBeSelf, memberNotRegistered([String])
```
nplasterer, line 9: Can you add the better error messaging here as well.

Sources/XMTPiOS/Conversations.swift
```diff
@@ -13,6 +18,77 @@ public actor Conversations {
...
+		let erroredAddresses = try await withThrowingTaskGroup(of: (String?).self) { group in
+			for address in addresses {
+				group.addTask {
+					if try await self.client.canMessageV3(address: address) {
+						return nil
+					} else {
+						return address
+					}
+				}
+			}
```
nplasterer, line 73: I like how you handled this. I can improve this on android for sure.

Sources/XMTPiOS/Group.swift
```diff
@@ -0,0 +1,119 @@
...
+
+	public static func == (lhs: Group, rhs: Group) -> Bool {
+		lhs.id == rhs.id
+	}
+
+	public func hash(into hasher: inout Hasher) {
+		id.hash(into: &hasher)
+	}
+
+	public var members: [String] {
```
nplasterer, line 39: Still think we should name this 
```suggestion
	public var memberAddresses: [String] {
```

Sources/XMTPiOS/Group.swift
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		let messages = try ffiGroup.findMessages(opts: options)
+
+		return try messages.map { ffiMessage in
+			let encodedContent = try EncodedContent(serializedData: ffiMessage.content)
+
+			return DecodedMessage(
+				client: client,
+				topic: "",
```
nplasterer, line 112: Been setting the topic as the group id since it's a unique identifier.

XMTP.podspec
```diff
@@ -44,5 +44,5 @@ Pod::Spec.new do |spec|
   spec.dependency "web3.swift"
   spec.dependency "GzipSwift"
   spec.dependency "Connect-Swift", "= 0.3.0"
-  spec.dependency 'LibXMTP', '= 0.4.1-beta0'
+  spec.dependency 'LibXMTP', '= 0.4.1-beta1'
```
nplasterer, line 47: This shouldnt be a change if you pull in main. 🤞

nplasterer (approved):
Looks like this PR is missing the call to list both groups and conversations in the same call. https://github.com/xmtp/xmtp-android/pull/160/files#diff-479fe53ececf1836cd910bfba9a721b20fce9fd212b204329abef95ff79429f9R220-R255
But probably okay to do that in a follow up.
Don't forget to bump the pod spec.

nakajima:
> Don't forget to bump the pod spec.

Do we think this should be a `0.9.0-beta`?

PR #230, opened 2024-02-02
Renaming Conversation -> DirectMessage

nakajima:
WIP

This renames `Conversation` to `DirectMessage`, which allows us to use `Conversation` as a union of Group and DirectMessage conversations.

Relies on https://github.com/xmtp/xmtp-ios/pull/229.


nplasterer:
Totally in favor of this but I don't think we should be doing any major breaking changes like this with out also giving them the groups functionality. Probably should wait till this is ready in Beta or Dev before merging.

PR #232, opened 2024-02-02
fix: set salt as empty

kele-leanes:
**Description**

This pull request updates the `hkdfHmacKey` method to set the salt as an empty `Data`, aligning it with the behavior observed in xmtp-js.

### DIFF
--- a/Sources/XMTPiOS/Crypto.swift
+++ b/Sources/XMTPiOS/Crypto.swift
@@ -8,7 +8,7 @@ import Foundation
 public typealias CipherText = Xmtp_MessageContents_Ciphertext
 
 enum CryptoError: Error {
-	case randomBytes, combinedPayload, keyDerivationError, hmacSignatureError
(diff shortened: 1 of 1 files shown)


Sources/XMTPiOS/Crypto.swift
```diff
@@ -105,17 +105,13 @@ enum Crypto {
...
-		} catch {
-			throw CryptoError.keyDerivationError
-		}
+		let salt = Data()
+		let key = HKDF<SHA256>.deriveKey(
+			inputKeyMaterial: SymmetricKey(data: secret),
+			salt: salt,
+			info: info,
+			outputByteCount: 32)
+		return key
```
nplasterer, line 114:
```suggestion
		let key = HKDF<SHA256>.deriveKey(
			inputKeyMaterial: SymmetricKey(data: secret),
			salt: Data(),
			info: info,
			outputByteCount: 32)
		return key
```

nplasterer (approved):

PR #233, opened 2024-02-05
Pull v3 client into its own PR

nakajima:
Contains the client related changes from https://github.com/xmtp/xmtp-ios/pull/229.

### DIFF
--- a/Sources/XMTPiOS/Client.swift
+++ b/Sources/XMTPiOS/Client.swift
@@ -22,10 +22,10 @@ public struct ClientOptions {
 		/// Specify which XMTP network to connect to. Defaults to ``.dev``
 		public var env: XMTPEnvironment = .dev
 
-		/// Optional: Specify self-reported version e.g. XMTPInbox/v1.0.0.
+		/// Specify whether the API client should use TLS security. In general this should only be false when using the `.local` environment.
 		public var isSecure: Bool = true
 
-		/// Specify whether the API client should use TLS security. In general this should only be false when using the `.local` environment.
+		/// /// Optional: Specify self-reported version e.g. XMTPInbox/v1.0.0.
 		public var appVersion: String?
 
 		public init(env: XMTPEnvironment = .dev, isSecure: Bool = true, appVersion: String? = nil) {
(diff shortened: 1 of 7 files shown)


Sources/XMTPiOS/Client.swift
```diff
@@ -95,28 +108,64 @@ public final class Client: Sendable {
...
-			throw ClientError.creationError(error.localizedDescription)
+			throw ClientError.creationError("\(error)")
+		}
+	}
+
+	static func initV3Client(address: String, options: ClientOptions?, source: LegacyIdentitySource, privateKeyBundleV1: PrivateKeyBundleV1) async throws -> FfiXmtpClient? {
+		let v3Client: FfiXmtpClient?
+
+		if options?.enableAlphaMLS == true && options?.api.env == .local {
+			let dbURL = URL.documentsDirectory.appendingPathComponent("xmtp-\(options?.api.env.rawValue ?? "")-\(address).db3")
```
nplasterer, line 119: I don't know how `rawValue` works but in Android this would return the actual value so 10.2.2.2 or whatever. I was thinking this would return just enum name like local, dev, or prod. Atleast thats how we do it in android. 🤷‍♀️ Not a huge deal tho.

Sources/XMTPiOS/Client.swift
```diff
@@ -172,13 +233,36 @@ public final class Client: Sendable {
...
+		)
+
+		if let v3Client, let textToSign = v3Client.textToSign() {
+			guard let signingKey else {
+				throw ClientError.creationError("No v3 keys found, you must pass a SigningKey in order to enable alpha MLS features")
+			}
+
+			let signature = try await signingKey.sign(message: textToSign)
+			try await v3Client.registerIdentity(recoverableWalletSignature: signature.rawData)
+		}
```
nplasterer, line 250: Since this check is required for V3 client creation I think it should move into the above method. And the above method should take an optional signer.

nplasterer: I also think the logic here is incorrect if there is no text to sign that just means a V3 identity already exists but you still need to call registerIdentity on it. If there is text to sign you sign it and call register identity. Otherwise then error.

```
        if (v3Client.textToSign() == null) {
            v3Client.registerIdentity(null)
        } else if (account != null) {
            v3Client.textToSign()?.let {
                v3Client.registerIdentity(account.sign(it))
            }
        } else {
            Log.i(TAG, "No signer passed but signer was required.")
        }
```

Sources/XMTPiOS/Client.swift
```diff
@@ -172,13 +233,36 @@ public final class Client: Sendable {
...
+
+		let result = try Client(address: address, privateKeyBundleV1: v1Bundle, apiClient: apiClient, v3Client: v3Client)
+
+		for codec in options.codecs {
+			result.register(codec: codec)
+		}
```
nplasterer, line 256: Is there a reason you added this here? Feels like if we didn't have it before than it doesn't make sense to add it now.

Tests/XMTPTests/ClientTests.swift
```diff
@@ -20,6 +20,38 @@ class ClientTests: XCTestCase {
...
+				)
+			)
+
+			XCTFail("did not throw")
+		} catch {
+			XCTAssert(true)
+		}
+	}
+
+
```
nplasterer, line 54: Might be nice to add a test for the creating with a bundle scenario since it's a bit of an outlier.

Sources/XMTPiOS/Client.swift
```diff
@@ -95,28 +108,64 @@ public final class Client: Sendable {
...
+			let signature = try await account.sign(message: textToSign).rawData
+			try await v3Client?.registerIdentity(recoverableWalletSignature: signature)
+		}
+
+		let client = try Client(address: account.address, privateKeyBundleV1: privateKeyBundleV1, apiClient: apiClient, v3Client: v3Client)
 		try await client.ensureUserContactPublished()
 
+		for codec in (options?.codecs ?? []) {
+			client.register(codec: codec)
+		}
```
nplasterer, line 157: Same about the codec code? Was this meant to be committed here?

nakajima: This was actually not working before.

Sources/XMTPiOS/Client.swift
nakajima: It was busted before.

Issue #185, opened 2023-11-08
Bug: Stream disconnects on App Backgrounding


fabriguespe:
**Issue:**
Streams intermittently stop receiving messages after the app is backgrounded and re-foregrounded.

**Reproduce:**
1. Initialize a stream
2. Background or the app.
3. Foreground the app.
4. No incoming messages received.

**Expected:**
Stream maintains connection or reconnects after foregrounding.

**Actual:**
Stream 'dies' silently without errors.

nplasterer:
It seems suspicious that this is only happening on backgrounding and then reforegrounding the app. Were you able to reproduce this @fabriguespe? I think until we can move the swift SDK over to the new `subscribe2` protocol function my advice would be to detect when network or background events happen and kill the stream and then restart the stream. Long term it would be good to do what we did here in Android https://github.com/xmtp/xmtp-android/pull/131 but this is a much heavier lift as it requires changes to `libxmtp`.

fabriguespe:
After running some tests, the `stream` fails silently after 4-10 minutes of the app being in the background.

I've implemented a solution addressing streams intermittently stopping after the app is backgrounded and re-foregrounded. The solution involves listening for specific app state notifications and managing the stream accordingly.

Here's the relevant code:

```swift
// Rest of the code ...
func startStream() {
  streamTask = Task {
	  do {
		  for try await message in conversation.streamMessages() {
			  let content: String = try message.content()
			  print("Received message: \(content)")
			  await MainActor.run {
				  messages.append(message)
			  }
		  }
	  } catch {
		  print("Error in message stream: \(error)")
	  }
     }	
}
// Rest of the code ...
.onReceive(NotificationCenter.default.publisher(for: UIApplication.willResignActiveNotification)) { in
 streamTask?.cancel()
}
.onReceive(NotificationCenter.default.publisher(for: UIApplication.willEnterForegroundNotification)) { in
  Task{
    await loadMessages()
    startStreamTask()
  }
}
```
In this code:

1. We listen for the `UIApplication.willResignActiveNotification` notification. When this notification is received, indicating the app is moving to the background, we cancel the `streamTask`, effectively stopping the stream.

2. We also listen for the `UIApplication.willEnterForegroundNotification` notification. When this notification is received, indicating the app is moving to the foreground, we create a new task that loads the messages and starts the stream again.

This approach ensures that the stream is properly managed when the app transitions between the foreground and background states, preventing it from 'dying' silently without errors.

PR #234, opened 2024-02-06
Update network links

nplasterer:
Fixes https://github.com/xmtp/xmtp-ios/issues/185

We had a small issue in how streaming was working in the backend. This should fix that.

### DIFF
--- a/Sources/XMTPiOS/ApiClient.swift
+++ b/Sources/XMTPiOS/ApiClient.swift
@@ -75,14 +75,6 @@ final class GRPCApiClient: ApiClient {
 		}
 	}
 
-	static func envToUrl(env: XMTPEnvironment) -> String {
-		switch env {
-		case XMTPEnvironment.local: return "http://localhost:5556"
-		case XMTPEnvironment.dev: return "https://grpc.dev.xmtp.network:443:5556"
-		case XMTPEnvironment.production: return "https://grpc.production.xmtp.network:443:5556"
-		}
-	}
-
 	func setAuthToken(_ token: String) {
 		authToken = token
 	}
--- a/Sources/XMTPiOS/ApiClient.swift
+++ b/Sources/XMTPiOS/ApiClient.swift
@@ -78,8 +78,8 @@ final class GRPCApiClient: ApiClient {
 	static func envToUrl(env: XMTPEnvironment) -> String {
 		switch env {
 		case XMTPEnvironment.local: return "http://localhost:5556"
-		case XMTPEnvironment.dev: return "https://dev.xmtp.network:5556"
-		case XMTPEnvironment.production: return "https://production.xmtp.network:5556"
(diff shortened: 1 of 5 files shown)


Sources/XMTPiOS/XMTPEnvironment.swift
```diff
@@ -9,7 +9,7 @@ import Foundation
 
 /// Contains hosts an `ApiClient` can connect to
 public enum XMTPEnvironment: String, Sendable {
-	case dev = "dev.xmtp.network",
-	     production = "production.xmtp.network",
+	case dev = "grpc.dev.xmtp.network:443",
```
neekolas, line 12: Are we sure this gets used? I think the more important place to update is [here](https://github.com/xmtp/xmtp-ios/blob/main/Sources/XMTPiOS/ApiClient.swift#L78)

nplasterer: Feels like a bug that we have these strings in two places. But just updated both for now. @nakajima might make sense to consolidate it into one in the future.

alexrisch (approved):

Sources/XMTPiOS/XMTPEnvironment.swift
alexrisch: It would be good to get them in the same place, doesn't need to happen in this, but maybe a scrub of our constants and putting in the same place

nakajima: Ya, we should drop https://github.com/xmtp/xmtp-ios/blob/main/Sources/XMTPiOS/ApiClient.swift#L78, it got introduced during the first round of Rust changes a while ago but isn't necessary.

Sources/XMTPiOS/ApiClient.swift
```diff
@@ -78,8 +78,8 @@ final class GRPCApiClient: ApiClient {
 	static func envToUrl(env: XMTPEnvironment) -> String {
 		switch env {
 		case XMTPEnvironment.local: return "http://localhost:5556"
-		case XMTPEnvironment.dev: return "https://dev.xmtp.network:5556"
-		case XMTPEnvironment.production: return "https://production.xmtp.network:5556"
+		case XMTPEnvironment.dev: return "https://grpc.dev.xmtp.network:443:5556"
+		case XMTPEnvironment.production: return "https://grpc.production.xmtp.network:443:5556"
```
nakajima, line 82: Should we have two port components on this URL?

nakajima (approved):

nplasterer (commented):
This looks good to me but I can't approve this. Can you bump the pod spec and document the breaking change/upgrade integrators will need to do?

Seems like just `GRPCApiClient.envToUrl(env: .local)` to  `XMTPEnvironment.local.url,` in their code.

PR #239, opened 2024-02-06
V3 group streaming

nakajima:
Relies on https://github.com/xmtp/xmtp-ios/pull/229.

### DIFF
--- a/Sources/XMTPiOS/Conversations.swift
+++ b/Sources/XMTPiOS/Conversations.swift
@@ -9,10 +9,25 @@ public enum GroupError: Error {
 	case alphaMLSNotEnabled, emptyCreation, memberCannotBeSelf, memberNotRegistered([String])
 }
 
+final class GroupStreamCallback: FfiConversationCallback {
+	let client: Client
+	let callback: (Group) -> Void
+
+	init(client: Client, callback: @escaping (Group) -> Void) {
+		self.client = client
+		self.callback = callback
+	}
+
+	func onConversation(conversation: FfiGroup) {
+		self.callback(conversation.fromFFI(client: client))
+	}
+}
+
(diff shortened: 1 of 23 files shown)


Sources/XMTPiOS/Conversations.swift
```diff
@@ -48,6 +63,18 @@ public actor Conversations {
 		return try await v3Client.conversations().list(opts: options).map { $0.fromFFI(client: client) }
 	}
 
+	public func streamGroups() async throws -> AsyncThrowingStream<Group, Error> {
```
nplasterer, line 66: I'm not sure how you're going to stream both groups and conversations at the same time. But to be able to do this in Android the stream had to return the same type.

Sources/XMTPiOS/Extensions/Ffi.swift
```diff
@@ -193,6 +193,22 @@ extension FfiV2SubscribeRequest {
...
 
+// MARK: Messages
+
+extension FfiMessage {
+	func fromFFI(client: Client) throws -> DecodedMessage {
+		let encodedContent = try EncodedContent(serializedData: content)
+
+		return DecodedMessage(
+			client: client,
+			topic: "",
```
nplasterer, line 204: If you can get the group id here that would be ideal.

nplasterer (commented):
This looks good maybe a bad merge with main because of all the env stuff in the diff?
Might be worth trying to tackle the more difficult item of streaming both conversations and groups at the same time.

Sources/XMTPiOS/Conversations.swift
nakajima: Currently planning on holding off on having a combined stream until we do the Conversation -> DM rename. Then we can just have that stream return the union.

Issue #221, opened 2024-01-24
Feature request: Bubble up specific API error codes to the application developer


neekolas:
### Is your feature request related to a problem?

When a developer hits an error in our API, they get back a very generic error (`Error: The operation couldn't be completed. (XMTP.ApiClientError error 1.)`. It would be helpful to know what type of error the node returned. Especially 429 errors from being rate limited

### Describe the solution to the problem

Modify the bindings in `libxmtp` to return more specific errors. Possibly modify the `grpc_api_client` to add more detail if needed. Then pass that along to the caller in `xmtp-ios`.

### Describe the uses cases for the feature

Detecting when you have been rate limited to back off from requests.

PR #240, opened 2024-02-06
Improve error messages

nakajima:
Should help out with https://github.com/xmtp/xmtp-react-native/issues/225.

Closes https://github.com/xmtp/xmtp-ios/issues/221.

Before:

> XMTPiOS.ConversationError error 2

After:

> Recipient cannot be sender


nplasterer (approved):
Can you bump the podspec.

PR #241, opened 2024-02-09
Update bindings

cameronvoell:


nplasterer (approved):
I think getting on the latest bindings makes sense. I'll bump the podspec and see if this fixes the cocoapod issues.

PR #246, opened 2024-02-10
Groups a sub-type of Conversation

nplasterer:
To support RN and bring iOS more inline with Android lets have groups inherit from the conversation


cameronvoell (approved):
looks good!

PR #248, opened 2024-02-13
Update group send to match android

nplasterer:
Try to bring the group send more inline with what android does to better the chances of decryption.


cameronvoell (approved):
LGTM!

PR #252, opened 2024-02-16
Fix v3 client creation for real devices

alexrisch:


Sources/XMTPiOS/Client.swift
```diff
@@ -170,11 +170,6 @@ public final class Client {
 			signingKey: account
 		)
 
-		if let textToSign = v3Client?.textToSign() {
```
neekolas, line 173: This seems important. Does it exist somewhere else?

nplasterer: It does it's just repeated code.

https://github.com/xmtp/xmtp-ios/blob/main/Sources/XMTPiOS/Client.swift#L143-L152

neekolas (approved):

nplasterer (approved):

PR #263, opened 2024-02-22
Fix peer addresses actually returns peer addresses

nplasterer:
Wasn't removing self correctly

### DIFF
--- a/Sources/XMTPiOS/Group.swift
+++ b/Sources/XMTPiOS/Group.swift
@@ -82,6 +82,14 @@ public struct Group: Identifiable, Equatable, Hashable {
 			return []
 		}
 	}
+	
+	public var peerAddresses: [String] {
+		var addresses = memberAddresses.map(\.localizedLowercase)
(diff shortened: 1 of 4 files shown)


cameronvoell (approved):

Sources/XMTPiOS/Group.swift
```diff
@@ -82,6 +82,14 @@ public struct Group: Identifiable, Equatable, Hashable {
 			return []
 		}
 	}
+	
+	public var peerAddresses: [String] {
```
tsachiherman, line 86: when reading this, it wasn't clear to me ( initially ) that this is the "other conversation members peer addresses" list. not sure if it's a good idea or not, but the `memberAddresses.map(\.localizedLowercase)` seems to be error prone, as it might create different address representation to the same address. Ideally, it would be lower-cased at the source.

nplasterer: That would be ideal but since we haven't been lowercasing from the beginning it's not that easy and would potentially break existing clients. Something we can strive for when we move to v3 tho.

PR #265, opened 2024-02-23
feat: get HMAC keys from conversations

kele-leanes:
**Description:**
This PR adds the `getHmacKeys` method to the Conversations class and its test as we currently have in JS and Android


richardhuaaa (approved):
This looks correct to me, and well done on implementing the `verify` for testing purposes.

I think the true test will be if clients on different platforms successfully interoperate - it may make sense to retrieve some hard-coded macs from each client and add a test case on the server that they all verify successfully. @rygine would you be able to coordinate that?

nplasterer (approved):
I think we also want to revert this PR right? https://github.com/xmtp/xmtp-ios/pull/231 and bump the pod spec so you can test it in RN.

kele-leanes:
yes, but in #231 there is a commit 8a02e8dc92f332462d77f8f3ec507f497bcc97dc that @zombieobject introduced. it can be also reverted?

nplasterer:
@kele-leanes yes fine to revert. @zombieobject can introduce that in a follow up PR 👍

PR #266, opened 2024-02-25
Delete database follows ups

nplasterer:
Adds a test and fixes the implementation a bit.


cameronvoell (approved):
delete db, new test look good 👍

PR #276, opened 2024-03-06
thirtyDayPeriodsSinceEpoch fix

nplasterer:
In JS they do this

```js
const thirtyDayPeriodsSinceEpoch = Math.floor(
          Date.now() / 1000 / 60 / 60 / 24 / 30
        )
```

I think this brings the code closer inline with that.

### DIFF
--- a/Sources/XMTPiOS/Conversations.swift
+++ b/Sources/XMTPiOS/Conversations.swift
@@ -587,7 +587,9 @@ public actor Conversations {
 	}
 	
 	public func getHmacKeys(request: Xmtp_KeystoreApi_V1_GetConversationHmacKeysRequest? = nil) -> Xmtp_KeystoreApi_V1_GetConversationHmacKeysResponse {
-		let thirtyDayPeriodsSinceEpoch = Int(Date().timeIntervalSince1970) / (60 * 60 * 24 * 30)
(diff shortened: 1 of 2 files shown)


Sources/XMTPiOS/Conversations.swift
```diff
@@ -587,7 +587,9 @@ public actor Conversations {
 	}
 	
 	public func getHmacKeys(request: Xmtp_KeystoreApi_V1_GetConversationHmacKeysRequest? = nil) -> Xmtp_KeystoreApi_V1_GetConversationHmacKeysResponse {
-		let thirtyDayPeriodsSinceEpoch = Int(Date().timeIntervalSince1970) / (60 * 60 * 24 * 30)
+		let daysSinceEpoch = Date().timeIntervalSince1970 / (60 * 60 * 24)
+		let thirtyDayPeriodsSinceEpoch = Int(floor(daysSinceEpoch / 30))
```
nplasterer, line 591: JS divides also by 1000 every where but I think that might be a bug in JS @rygine

kele-leanes (approved):

nplasterer:
Confirmed that iOS convert to int takes the floor so I actually don't think we need this.

PR #280, opened 2024-03-11
Fix list conversation only listing first 100

nplasterer:
This validates that the fixes in libxmtp now fix the pagination issue for listing all conversations

Also adds a test around performance. Benchmarks that conversation listing off a large wallet should take less than 30 seconds. A few results showed:
`Loaded 2015 conversations in 17.3847593298321s`
`Loaded 2015 conversations in 8.91787302494049s`

### DIFF
--- a/Tests/XMTPTests/ConversationsTest.swift
+++ b/Tests/XMTPTests/ConversationsTest.swift
@@ -194,7 +194,7 @@ class ConversationsTests: XCTestCase {
 		}
 	}
 	
-	func testLoadConvos() async throws {
+	func testLoadAllConvosForLargeWallet() async throws {
 		// Data from hex: 0836200ffafa17a3cb8b54f22d6afa60b13da48726543241adc5c250dbb0e0cd
 		// aka 2k many convo test wallet
 		let privateKeyData = Data([8,54,32,15,250,250,23,163,203,139,84,242,45,106,250,96,177,61,164,135,38,84,50,65,173,197,194,80,219,176,224,205])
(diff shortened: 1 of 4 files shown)


alexrisch (approved):

Tests/XMTPTests/ConversationsTest.swift
```diff
@@ -193,4 +193,21 @@ class ConversationsTests: XCTestCase {
 			}
 		}
 	}
+	
+	func testLoadAllConvosForLargeWallet() async throws {
+		// Data from hex: 0836200ffafa17a3cb8b54f22d6afa60b13da48726543241adc5c250dbb0e0cd
+		// aka 2k many convo test wallet
+		let privateKeyData = Data([8,54,32,15,250,250,23,163,203,139,84,242,45,106,250,96,177,61,164,135,38,84,50,65,173,197,194,80,219,176,224,205])
```
neekolas, line 200: One worry about including this PK in the repo is that random devs are going to use it and start adding more convos to the account, or even misusing it to accidentally have broken convos. That makes it less useful as a testing tool.

neekolas (approved):

Tests/XMTPTests/ConversationsTest.swift
nplasterer: Ya I've been hesitant to commit any of these for awhile. But they keep creeping up with issues. I'll remove it for now. A good test that creates 101 convos may suffice. Atleast for pagination if not for performance.

neekolas: ➕ on a 101 conversation test

PR #281, opened 2024-03-12
Fix Homebrew + Python CI Error

zombieobject:
## Introduction 📟
A Homebrew formula that depended on a Python migration was failing with the following error:

```
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3
```

## Purpose ℹ️ 
The purpose of this pull request is to add a "temporary" workaround until a fix is introduced into the Homebrew pipeline by first updating Homebrew and then cleaning out the install directory for `python 2to3`.  The automation then directs the GitHub action to install Colima and let any sub steps fail gracefully; after which the python link is overwritten.

## Scope 🔭
- Github Integrations Test Workflow

## Discussion 🎙
Once the fix for this issue enters the Homebrew pipeline, the workflow will be updated.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ env:
 jobs:  
   ios-tests:
     name: iOS Tests
-    # macos-14 currently breaks colima
+    # macos-14 currently breaks Colima
     runs-on: macos-13
     
     steps:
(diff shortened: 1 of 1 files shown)


neekolas (approved):

.github/workflows/tests.yml
```diff
@@ -22,13 +22,53 @@ jobs:
...
-
+        
+        # This step and the Install Colima step have been modified to address a
+        # Python error started 3.11.2024, emitting the following error:
+        #
+        # Error: The `brew link` step did not complete successfully
+        # The formula built, but is not symlinked into /usr/local
+        # Could not symlink bin/2to3
+        #
+        # Others reported: https://github.com/Cockatrice/Cockatrice/issues/5007
```
alexrisch, line 33: Thank you for linking the issue 👍

PR #305, opened 2024-04-03
Fix lazy load of conversation class

nplasterer:
We are seeing a bug in React Native outlined here: https://github.com/xmtp/xmtp-react-native/issues/343

It appears on call to conversations the promise can execute 10 times create a new conversations class each time before reusing the same one. I think this might be because of the lazy load of conversations which waits to create the class until it's first called. Seeing as the fix was to just call conversations once before calling the promise I think thats the case.

I'm not too familiar with how iOS works but it doesnt appear I can init the conversations class in the Client init because it takes a client as a parameter. I've opted to just calling conversations in the create function to short circuit the lazy and load it on create.


nplasterer:
Went deep on this with @zombieobject again and this does seems like the best solution given the constraints of how the class is setup. May be a future refactor we can clean this up.

PR #307, opened 2024-04-04
feat: Pagination on Consent Load

alexrisch:
Added optional pagination to consent list loading

### DIFF
--- a/Sources/XMTPiOS/Contacts.swift
+++ b/Sources/XMTPiOS/Contacts.swift
@@ -45,7 +45,7 @@ public class ConsentList {
 	var publicKey: Data
 	var privateKey: Data
 	var identifier: String?
-
+  var lastFetched: Date?
 	var client: Client
 
 	init(client: Client) {
@@ -55,17 +55,18 @@ public class ConsentList {
 		identifier = try? LibXMTP.generatePrivatePreferencesTopicIdentifier(privateKey: privateKey)
 	}
 
-	func load(afterDate: Date? = nil) async throws -> ConsentList {
+	func load() async throws -> [ConsentListEntry] {
 		guard let identifier = identifier else {
 			throw ContactError.invalidIdentifier
 		}
+		let newDate = Date()
 
 		let pagination = Pagination(
-      after: afterDate,
-      direction: .ascending
-    )
+            after: lastFetched,
+            direction: .ascending
+        )
 		let envelopes = try await client.apiClient.envelopes(topic: Topic.preferenceList(identifier).description, pagination: pagination)
-		let consentList = ConsentList(client: client)
+    lastFetched = newDate
 
 		var preferences: [PrivatePreferencesAction] = []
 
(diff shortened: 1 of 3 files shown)


nplasterer:
We could make the 2k wallet on the dev network have a bunch of consent records to test this as well?

alexrisch:
> We could make the 2k wallet on the dev network have a bunch of consent records to test this as well?

We can, was working with Daria and she had a similar wallet to the one seeing a problem and it didn't seem specific to that use case. However it could be ideal to make the 2k wallet have additional consent records for testing this and verifying performance in the future

alexrisch:
I've added a couple thousand consent entries on to the dev wallet, this will make testing a little more real in the future

Sources/XMTPiOS/Contacts.swift
```diff
@@ -55,12 +55,13 @@ public class ConsentList {
...
 	}
 
-	func load() async throws -> ConsentList {
+	func load(pagination: Pagination? = nil) async throws -> ConsentList {
 		guard let identifier = identifier else {
 			throw ContactError.invalidIdentifier
 		}
 
-		let envelopes = try await client.apiClient.envelopes(topic: Topic.preferenceList(identifier).description, pagination: Pagination(direction: .ascending))
+		let pagination = pagination ?? Pagination(direction: .ascending)
```
nplasterer, line 63: I think we have to force whatever they pass to have the direction of ascending.... Otherwise the list will load backwards and it would be the opposite of what the actual consent is. We probably need to protect against that.

Tests/XMTPTests/ContactsTests.swift
```diff
@@ -81,4 +81,19 @@ class ContactsTests: XCTestCase {
...
+    
+    func testRefreshConsentListPagination() async throws {
+        let fixtures = await fixtures()
+        let contacts = fixtures.bobClient.contacts
+        let aliceAddress = fixtures.alice.address
+        try await contacts.deny(addresses: [aliceAddress])
+        let date = Date()
+        try await contacts.allow(addresses: [aliceAddress])
+
+        let result = try await contacts.consentList.load(pagination: Pagination(before: date, direction: .ascending))
```
nplasterer, line 93: So load is a private function so external users can't call it. They call it via refreshConsentList. So I actually think you need to add pagination parameters to the refreshConsentList as well.

nplasterer: I think we need to be strategic about this. We really only should accept a start date. Because we want to make sure the fetch all the latest and not just a couple in the middle since things could change. And we need it to always have the direction of ascending.

alexrisch: Personal thought was that if they want to paginate in whatever way they can, but most if not all use cases would be ascending
However I can update to be safer for the SDK usage

neekolas: Agree with @nplasterer. Going in descending order is guaranteed to mangle your consent list state. Can't see why we'd offer that kind of footgun

Sources/XMTPiOS/Contacts.swift
```diff
@@ -55,12 +55,13 @@ public class ConsentList {
...
+		let envelopes = try await client.apiClient.envelopes(topic: Topic.preferenceList(identifier).description, pagination: pagination)
```
neekolas, line 64: Isn't this already paginated under the hood?

What problem are we solving by putting the user in control of pagination here? When would they want to do anything other than load the entire consent list? Anything less would leave them in an out of date state.

Tests/XMTPTests/ContactsTests.swift
alexrisch: @neekolas @nplasterer Updated to just be after date so integrator can pass in their latest cached consent date

Sources/XMTPiOS/Contacts.swift
alexrisch: Updated to only take an after date to allow for more performant local cache updates

nplasterer: I think there needs to be a bigger change here because the refreshConsentList just wipes out the current list and makes a new list. I think the ideal scenario is that the `consentList` doesn't get reset. On every load and that you could just fetch the latest consent items from the last time that you fetched consent. But I think you'll need more changes for that.

alexrisch: This is kinda my issue with current consent, I think it's nuanced (at least personally)

alexrisch: An integrator would only use afterDate if they were caching the consent list, so they likely would not be using the SDK memory cached consent list after refreshing

neekolas: > An integrator would only use afterDate if they were caching the consent list

Ahh. I see. But our SDK also needs an up-to-date consent list, doesn't it? Is the idea that the integrator would pass their known entries back in to us from their cache on startup?

Alternatively, we could just handle caching for the integrator. I know on iOS there's all kinds of permissions implications, but maybe it's worthwhile anyways.

Sources/XMTPiOS/Contacts.swift
```diff
@@ -70,75 +75,78 @@ public class ConsentList {
...
 			}
 
 			for groupId in preference.denyGroup.groupIds {
-				_ = consentList.denyGroup(groupId: groupId)
+				_ = denyGroup(groupId: groupId)
 			}
 		}
 
-		return consentList
+        return Array(entries.values)
```
nplasterer, line 96: What does `entries.values` return if not an array. 🤔

Sources/XMTPiOS/Contacts.swift
```diff
@@ -70,75 +75,78 @@ public class ConsentList {
...
+    )
+
+    let envelope = Envelope(
+        topic: Topic.preferenceList(identifier),
+        timestamp: Date(),
+        message: Data(message)
+    )
+
+    try await client.publish(envelopes: [envelope])
+  }
```
nplasterer, line 149: I think this logic is a little incorrect. I think we want to pass in an array of envelopes. Like this where the envelopes are returned from the loop.

```suggestion
    func publish(entries: [ConsentListEntry]) async throws {
      guard let identifier = identifier else {
        throw ContactError.invalidIdentifier
      }
      var payload = PrivatePreferencesAction()

      let envelopes = entries.map { entry ->
        switch entry.entryType {

    	    case .address:
    	      switch entry.consentType {
    	      case .allowed:
    	          payload.allowAddress.walletAddresses.append(entry.value)
    	      case .denied:
    	          payload.denyAddress.walletAddresses.append(entry.value)
    	      case .unknown:
    	          payload.messageType = nil
    	      }

    			case .groupId:
    	    	switch entry.consentType {
    	    	case .allowed:
    	    	    if let valueData = entry.value.data(using: .utf8) {
    	    	        payload.allowGroup.groupIds.append(valueData)
    	    	    }
    	    	case .denied:
    	    	    if let valueData = entry.value.data(using: .utf8) {
    	    	            payload.denyGroup.groupIds.append(valueData)
    	    	    }
    	    	case .unknown:
    	    	    payload.messageType = nil
    	    }
    	}
    	
        let message = try LibXMTP.userPreferencesEncrypt(
          publicKey: publicKey,
          privateKey: privateKey,
          message: payload.serializedData()
        )

        Envelope(
          topic: Topic.preferenceList(identifier),
          timestamp: Date(),
          message: Data(message)
         )
    }

    try await client.publish(envelopes: envelopes)
  }
```

Sources/XMTPiOS/Contacts.swift
```diff
@@ -196,6 +204,7 @@ public actor Contacts {
 
 	// Whether or not we have sent invite/intro to this contact
 	var hasIntroduced: [String: Bool] = [:]
+    var latestUpdate: Date?
```
nplasterer, line 207: Do you need this? Probably shouldn't have two sources of truth for the last time it was updated.

Sources/XMTPiOS/Contacts.swift
alexrisch: @nplasterer I can update to multiple envelopes, but this payload supports multiple addresses being added (not sure which is actually ideal from network/device performance)

nplasterer: I'm not sure I understand the question?

nplasterer: Ooo you're using append instead of add. 🤔

alexrisch: > Ooo you're using append instead of add. 🤔

Swift :)

> I'm not sure if we tested this case. I think if you just add a test for calling `deny(addresses: [String]) ` with a list of addresses. And make sure that all the addresses are denied. This should be fine with just one envelope. 🤔

Added a new test case to verify those

nplasterer (approved):
Approving to unblock. Don't forget to bump the Pod file so you can make a release.

PR #326, opened 2024-04-24
Update deploy_to_cocoapods.yml

nplasterer:
The release is failing looks like an update to latest recently happened.


cameronvoell (approved):
worth a shot

PR #333, opened 2024-05-07
fix: Finally fix Consent Proof iOS

alexrisch:
Fix one letter


nplasterer (approved):
Don't forget to bump the podspec

PR #339, opened 2024-05-23
Fix SQL App Group Crash

nplasterer:
Should close https://github.com/xmtp/xmtp-ios/issues/336

This adds two new functions that should be called on background and foreground of the app when using app groups with the local database to disconnect and reconnect the database to avoid crashing.

### DIFF
--- a/Sources/XMTPiOS/Client.swift
+++ b/Sources/XMTPiOS/Client.swift
@@ -462,6 +462,21 @@ public final class Client {
 		let fm = FileManager.default
 		try fm.removeItem(atPath: dbPath)
 	}
+	
+	@available(*, deprecated, message: "This function is delicate and should be used with caution. App will error if database not properly reconnected. See: reconnectLocalDatabase()")
+	public func dropLocalDatabaseConnection() throws {
+		guard let client = v3Client else {
+			throw ClientError.noV3Client("Error no V3 client initialized")
+		}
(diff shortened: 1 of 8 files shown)


Sources/XMTPiOS/Client.swift
```diff
@@ -462,6 +462,21 @@ public final class Client {
 		let fm = FileManager.default
 		try fm.removeItem(atPath: dbPath)
 	}
+	
+	@available(*, deprecated, message: "This function is delicate and should be used with caution. App will error if database not properly reconnected. See: reconnectLocalDatabase()")
```
nplasterer, line 466: @insipx was able to mark this function as delicate to warn developers before using.

Sources/XMTPiOS/Client.swift
```diff
@@ -462,6 +462,21 @@ public final class Client {
...
+	public func dropLocalDatabaseConnection() throws {
+		guard let client = v3Client else {
+			throw ClientError.noV3Client("Error no V3 client initialized")
+		}
+		try client.releaseDbConnection()
+	}
+	
+	public func reconnectLocalDatabase() async throws {
```
nplasterer, line 474: Should be any reason to mark this as delicate because if they try to reconnect a already connected database nothing bad should happen right?

cameronvoell (approved):

PR #343, opened 2024-05-30
Inconsistency with timestamps

nplasterer:
Fixes the date issue xmtp/xmtp-react-native#394 and the membership list.

Also makes some better naming for consent items


cameronvoell (approved):
looks good, thanks for removing my comments from test as well 👍

PR #358, opened 2024-06-28
V2 Streaming upgrades

nplasterer:
Actually use subscribe2 correctly without infinite while loop.

### DIFF
--- a/Sources/XMTPTestHelpers/TestHelpers.swift
+++ b/Sources/XMTPTestHelpers/TestHelpers.swift
@@ -67,208 +67,20 @@ public struct FakeWallet: SigningKey {
 	}
 }
 
-enum FakeApiClientError: String, Error {
-	case noResponses, queryAssertionFailure
-}
-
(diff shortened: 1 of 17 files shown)


Sources/XMTPTestHelpers/TestHelpers.swift
```diff
@@ -264,15 +259,19 @@ public struct Fixtures {
...
 	public var bobClient: Client!
+	let clientOptions: ClientOptions? = ClientOptions(
+		api: ClientOptions.Api(env: XMTPEnvironment.local, isSecure: false)
+	)
 
 	init() async throws {
 		alice = try PrivateKey.generate()
 		bob = try PrivateKey.generate()
 
 		fakeApiClient = FakeApiClient()
```
nplasterer, line 270: I really want to remove all the fakeApiClient stuff from here too. A lot of tests aren't really testing what we think they are.

cameronvoell (approved):
nice, tests looking good!

PR #359, opened 2024-07-01
Change group consent to string instead of bytes

nplasterer:
Closes https://github.com/xmtp/xmtp-ios/issues/357

Changes this to a string to help with translation issues.


cameronvoell (approved):
👍

Issue #368, opened 2024-07-22
Improve Swift error messaging in RN


nplasterer:
In the reaact native logs I only see : ERROR in create:  [Error: The operation couldn't be completed. (XMTP.ClientError error 0.)] which is really not readable
When building through Xcode I can see more logs: 'ERROR in create: ', { [Error: The operation couldn't be completed. (XMTP.ClientError error 0.)] code: 'ERR_UNEXPECTED' } but still nothing clear
This error is thrown here : https://github.com/xmtp/xmtp-ios/blob/9ed0edf5be557b01fd55b6d94a2883ffcbd7b4b4/Sources/XMTPiOS/Client.swift#L131
If I add a print("ERROR \(error)") above this line I do get : ERROR Storage(message: "Storage error: The SQLCipher Sqlite extension is not present, but an encryption key is given") so libxmtp does throw the right error but it’s not well surfaced in swift logs & RN logs. I’m adding the right linking to Converse to fix the issue but just know that for other devs it might really not be clear what the issue is

PR #370, opened 2024-07-23
Make the error messages readable

nplasterer:
Closes https://github.com/xmtp/xmtp-ios/issues/368

| Before | After |
|--------|--------|
| Screenshot 2024-07-23 at 5 30 50 PM | Screenshot 2024-07-23 at 5 28 50 PM |


nplasterer:
I'm sure there is more places we can improve these error messages but lets start here. As we run into other ones we can add them case by case using this template.

alexrisch (approved):

PR #372, opened 2024-07-24
Update visibility of PermissionPolicy fields

cameronvoell:

### DIFF
--- a/Tests/XMTPTests/GroupPermissionsTests.swift
+++ b/Tests/XMTPTests/GroupPermissionsTests.swift
@@ -7,7 +7,7 @@
 
 import CryptoKit
 import XCTest
-@testable import XMTPiOS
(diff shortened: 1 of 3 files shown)


Tests/XMTPTests/GroupPermissionsTests.swift
```diff
@@ -7,24 +7,46 @@
 
 import CryptoKit
 import XCTest
-@testable import XMTPiOS
```
cameronvoell, line 10: Created an issue for removing `@testable` from other tests unless its explicitly needed. This will help us catch inaccessible fields in the future and keep us from using correct fields in tests that arent available to SDK users, like how this test was previously using internal only `walletAddress` field instead of the public `address` field.

nplasterer (approved):

PR #381, opened 2024-08-07
Adds default message sync url

nplasterer:
This adds a default message sync url and bumps to the latest rust.

### DIFF
--- a/Sources/XMTPiOS/Client.swift
+++ b/Sources/XMTPiOS/Client.swift
@@ -75,7 +75,7 @@ public struct ClientOptions {
 		enableV3: Bool = false,
 		encryptionKey: Data? = nil,
 		dbDirectory: String? = nil,
-		historySyncUrl: String? = nil
+		historySyncUrl: String = "https://message-history.dev.ephemera.network/"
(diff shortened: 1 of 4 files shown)


Sources/XMTPiOS/Client.swift
```diff
@@ -75,7 +75,7 @@ public struct ClientOptions {
 		enableV3: Bool = false,
 		encryptionKey: Data? = nil,
 		dbDirectory: String? = nil,
-		historySyncUrl: String? = nil
+		historySyncUrl: String = "https://message-history.dev.ephemera.network/"
```
nplasterer, line 78: @neekolas just confirming this is the url we're using now.

nplasterer: Is there a prod network as well? Should this correlate with the env the user is using? FYI @jhaaaa

neekolas: Yes, that’s the right URL. There is a production version deployed as well at https://message-history.production.ephemera.network/

alexrisch (approved):

PR #383, opened 2024-08-08
Add additional wallets to a client

nplasterer:
This adds the ability to add additional wallets to an existing client.

Client creation only allows creating with an EOA so we can guarantee V2 functionality works as intended. However you can add an additional EOA or SCW.

On initial client creation a chainRPC needs to be set and on all additional SCW added they need to conform to CAIP addressing https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md


Sources/XMTPiOS/SigningKey.swift
```diff
@@ -19,6 +19,9 @@ import LibXMTP
 public protocol SigningKey {
 	/// A wallet address for this key
 	var address: String { get }
+	
+	/// Chain rpc url for the smart contract wallet
+	var chainRPCUrl: String? { get }
```
nplasterer, line 24: @neekolas I think I need some knowledge sharing on how this chainRPC is supposed to be set. If it's okay for it to be on the signing key which is an interface over written by the integrator.

Sources/XMTPiOS/Client.swift
```diff
@@ -602,4 +602,26 @@ public final class Client {
 		}
 		try await client.requestHistorySync()
 	}
+	
+	public func addWallet(account: SigningKey) async throws {
```
nplasterer, line 606: After adding the wallet to their identity what benefits do they get/ how can the interact with the wallet?

Sources/XMTPiOS/SigningKey.swift
neekolas: This was all done a while ago, and not by me, so my recollection is a little fuzzy on where and when we need to pass in the chainRPC.

But generally it's a thing that dapps will have (it's the URL you'd get from an Infura or Alchemy dashboard). It wouldn't be "key specific", so it's probably safe to make it a client create parameter and save it there. We'd use the same RPC url for all signing keys.

We need it at the time we are creating or verifying smart contract wallet signatures.

Sources/XMTPiOS/Client.swift
neekolas: This would allow someone else to enter any one of the added wallet addresses and reach the same inbox.

And for an app that has access to any of the added wallets to create a new installation for that inbox.

Sources/XMTPiOS/Client.swift
```diff
@@ -602,4 +602,26 @@ public final class Client {
...
+			throw ClientError.noV3Client("Error: No V3 client initialized")
+		}
+
+		do {
+			let signatureRequest = try await client.addWallet(existingWalletAddress: self.address, newWalletAddress: account.address)
+			
+			let signedData = try await account.sign(message: signatureRequest.signatureText())
+			
+			if let chainRPCUrl = account.chainRPCUrl {
+				try await signatureRequest.addScwSignature(signatureBytes: signedData.rawData, address: account.address, chainRpcUrl: chainRPCUrl)
```
nplasterer, line 617: Link this here as this is how we will want to send the address to the signature request https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md

Tests/XMTPTests/ClientTests.swift
```diff
@@ -455,4 +455,73 @@ class ClientTests: XCTestCase {
...
+		XCTAssertEqual(alix.addresses.count, 1)
+
+		let alixWallet2 = try FakeSCWWallet.generate()
+		try await alix.addWallet(account: alixWallet2)
+		XCTAssertEqual(alix.addresses.count, 2)
+	}
+	
+	public struct FakeSCWWallet: SigningKey {
+		public static func generate() throws -> FakeWallet {
+			let key = try PrivateKey.generate()
```
nplasterer, line 498: I don't know if there is something simple like this for generating a test SCW... seems unlikely. But I can atleast test the account format and isSmartContractWallet.

Tests/XMTPTests/ClientTests.swift
```diff
@@ -455,4 +455,42 @@ class ClientTests: XCTestCase {
...
+				api: .init(env: .local, isSecure: false),
+				   enableV3: true,
+				   encryptionKey: key
+			   )
+			)
+		XCTAssertEqual(alix.addresses.count, 1)
+
+		let alixWallet2 = try PrivateKey.generate()
+		try await alix.addWallet(account: alixWallet2)
+		XCTAssertEqual(alix.addresses.count, 2)
```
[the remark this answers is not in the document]
neekolas: The one we already expose is `group.members()` which will show all the account addresses for each member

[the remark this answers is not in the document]
nplasterer: I can for sure use that in the test. But what if I want to see all the addresses I have on my account so I can revoke some? Should we also expose addresses on client?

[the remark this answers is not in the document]
neekolas: Easy enough to add. We have all the information in the local DB. Happy to pick that one up, should be a small task.

Will be helpful for building out some of the user flows around revocation too, where you need to see all your own account addresses and installations.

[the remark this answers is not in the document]
nplasterer: When we talk about revoking a wallet that is different than revoking installations correct? Revoking a wallet would just revoke all the installations for that address?

Sources/XMTPiOS/SigningKey.swift
```diff
@@ -19,6 +19,9 @@ import LibXMTP
 public protocol SigningKey {
 	/// A wallet address for this key
 	var address: String { get }
+	
+	/// If this signing key is a smart contract wallet
+	var isSmartContractWallet: Bool { get }
```
nplasterer, line 24: Also pass a chainId

PR #386, opened 2024-08-16
Fix the thread locking in swift

nplasterer:
Reproduce tests from here: https://github.com/xmtp/xmtp-react-native/pull/470/files

Previously this would take over 60 seconds. Now takes under a second.


nplasterer:
I actually think this doesn't reproduce. Likely a RN threading issue again.

nplasterer:
Okay able to reproduce the issue but only with the members list.

insipx (approved):

PR #390, opened 2024-08-21
Added 2 tests showing a database lock when syncing many groups or one group many times

nmalzieu:
2 groups showing something we experience in RN

If I have many (40 ) groups and try to sync them all, I get a `Pool error: timed out waiting for connection` 
I can also reproduce by calling sync on a single group but many times

After 30 secs: 

Capture d’écran 2024-08-21 à 18 09 37

### DIFF
--- a/Tests/XMTPTests/GroupTests.swift
+++ b/Tests/XMTPTests/GroupTests.swift
@@ -858,13 +858,67 @@ class GroupTests: XCTestCase {
 		}
 	}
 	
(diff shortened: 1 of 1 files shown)


nplasterer:
Andrew and I merged some fixes but it still didn't fix these tests. I think for the time being you should switch the `syncAllGroups` method which seems to sync over 100 groups in less than 1 second.

PR #395, opened 2024-09-02
Fix stream cancelation cancels underlying stream

nplasterer:
Even after a stream was canceled the underlying stream was still streaming messages which was causing issues.

This officially cancels the stream and improves streams so only the stream messages are streamed once and not streamed when canceled

### DIFF
--- a/Sources/XMTPiOS/Conversations.swift
+++ b/Sources/XMTPiOS/Conversations.swift
@@ -426,20 +426,26 @@ public actor Conversations {
 
 	public func streamAllGroupDecryptedMessages() -> AsyncThrowingStream<DecryptedMessage, Error> {
 		AsyncThrowingStream { continuation in
-			Task {
-				do {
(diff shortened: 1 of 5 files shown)


nplasterer:
I'm actually getting this error alot in a bunch of random tests

> failed: caught error: "ClientError.creationError: LibXMTP.GenericError.Storage(message: "Storage error: Pool error: timed out waiting for connection: Unable to open the database file")"

I don't think it's related to these changes but figured I'd flag it. Wonder if anyone else is seeing these when running the tests in XCode?

nplasterer:
I'm no longer seeing the double send for messages either. 🎉 
So I think that means this is longer an issue
https://github.com/xmtp/libxmtp/issues/1023

alexrisch (approved):

alexrisch (approved):

neekolas (approved):

PR #402, opened 2024-09-20
Consent V3

nplasterer:
Part of https://github.com/xmtp/libxmtp/issues/801

This adds the ability to get and set consent records in V3

Dual sends so we are always setting both V2 and V3 consent. And adds a check to see if a V2Client is present to support pure V3 clients.


cameronvoell (approved):
looks good!

PR #403, opened 2024-09-20
Smart Contract Wallets

nplasterer:
Create an account with a smart contract wallet.

### DIFF
--- a/Sources/XMTPiOS/Client.swift
+++ b/Sources/XMTPiOS/Client.swift
@@ -14,6 +14,7 @@ public typealias PreEventCallback = () async throws -> Void
 public enum ClientError: Error, CustomStringConvertible, LocalizedError {
 	case creationError(String)
 	case noV3Client(String)
+	case noV2Client(String)
 
 	public var description: String {
 		switch self {
@@ -21,6 +22,8 @@ public enum ClientError: Error, CustomStringConvertible, LocalizedError {
 			return "ClientError.creationError: \(err)"
 		case .noV3Client(let err):
 			return "ClientError.noV3Client: \(err)"
+		case .noV2Client(let err):
+			return "ClientError.noV2Client: \(err)"
 		}
 	}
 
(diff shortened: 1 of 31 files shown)


Sources/XMTPiOS/Client.swift
```diff
@@ -696,4 +703,22 @@ public final class Client {
 		}
 		return InboxState(ffiInboxState: try await client.inboxState(refreshFromNetwork: refreshFromNetwork))
 	}
+	
+	// See for more details https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md
+	public func isValidAccountID(_ accountAddress: String) -> Bool {
```
nplasterer, line 708: @neekolas Is this the expected pattern for the address of a SCW?

Sources/XMTPiOS/SigningKey.swift
```diff
@@ -29,6 +38,18 @@ public protocol SigningKey {
...
+		return false
+	}
+	
+	public var chainId: UInt64 {
+		return "eip155:1"
+	}
+	
+	public var blockNumber: UInt64 {
+		return "eip155:1"
+	}
```
nplasterer, line 51: Could you provide me with what the default of these would be? @neekolas

Sources/XMTPiOS/Client.swift
neekolas: Yup, SCWs are all identified in CAIP-10 format

Sources/XMTPiOS/SigningKey.swift
neekolas: There is no default block number. I guess we can set it as 0 for EOAs.

There's two ways to go about getting the block number: we can make it the caller's problem and have them provide it at the time they are providing the signature (any eth client should be able to trivially look it up), or we can figure it out for them. Right now it's configured to happen the first way, but we might want to do it the second way because it's easier.

nplasterer: The chainId is also UInt64? Do you want me to encrypt it before passing? Is that the expected format?

neekolas: No need to encrypt. It’s totally public

Tests/XMTPTests/ClientTests.swift
```diff
@@ -514,4 +521,28 @@ class ClientTests: XCTestCase {
 		let newState = try await alixClient3.inboxState(refreshFromNetwork: true)
 		XCTAssertEqual(newState.installations.count, 1)
 	}
+	
+	func testCreatesASCWClient() async throws {
+		throw XCTSkip("TODO: Need to write a SCW local deploy with anvil")
```
nplasterer, line 526: I spent a bunch of time trying to get a SCW running locally like I did in Android. But I'm just not as familiar with the tooling in iOS and I time boxed it. Will revisit when I have more time. Will use the example in RN to validate.

codabrink (approved):

PR #406, opened 2024-09-26
Fixing group streams

nmalzieu:
## Introduction 📟
Sometimes groups were not getting streamed well in the RN SDK, especially when streaming messages at the same time

## Purpose ℹ️ 
Make group streaming still work even if streaming all messages at the same time

## Scope 🔭
We were using a single `StreamHolder` so when cancelling a `Task` if something was streamed we were detecting it (with for instance `Task.isCancelled` and calling `self.streamHolder.stream?.end()` on a stream which might not be the one we thought! Using an actor inside each streaming method to save & end the stream fixes it.

## Testing 🧪

I never managed to do a reproducible test directly in Swift. However I do have a failing test in the React Native repo: `can stream groups and messages` is failing on iOS on my side, but if I bring these Swift changes inside the XMTP Pod, it passes.

See https://github.com/xmtp/xmtp-react-native/pull/503

### DIFF
--- a/Sources/XMTPiOS/Conversations.swift
+++ b/Sources/XMTPiOS/Conversations.swift
@@ -87,11 +87,22 @@ class StreamManager {
 	}
 }
 
+actor FfiStreamActor {
+    private var ffiStream: FfiStreamCloser?
+
+    func setFfiStream(_ stream: FfiStreamCloser?) {
+        ffiStream = stream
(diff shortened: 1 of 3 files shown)


nplasterer (approved):
Nice find! Probably need to bump the pod so we can get it in RN.

PR #416, opened 2024-11-05
Update deploy_to_cocoapods.yml

nplasterer:
macos-12 is no longer supported


nplasterer:
I cant get these actions to work and I'm going to wait till we remove these dependencies before the full deprecation in December https://github.com/actions/runner-images/issues/10721

PR #426, opened 2024-11-13
List of inbox states for inbox ids

nplasterer:
Exposes the inbox state for inbox ids so that you can map addresses to a list of inbox ids.

Also adds a function to findDmByInboxId to make it easier for clients not storing addreses.


mchenani (approved):
🧹

PR #473, opened 2025-02-25
Granular Control Over Creation Flow

nplasterer:
Exposes Ffi functions that allow for more granular control over client creation so signing and registering can be handled.

All of this is delicate because takes advanced understanding of the protocol to use correctly.


cameronvoell (approved):
test and updates look good 👍

PR #476, opened 2025-03-07
Passkey Read

nplasterer:
- [x]  Make sure addWallet is blocking reassignment for the ffi function as well
- [x]  Adds new identity for all places where addresses were used
- [x]  Changes addMember, removeMember, createGroup, and newConversation to default to InboxId
- [x]  Enforces type on inboxId so address can't accidentally be passed.
- [x] Add a test to verify that an error gets thrown if an address is passed

### DIFF
--- a/Sources/XMTPiOS/Client.swift
+++ b/Sources/XMTPiOS/Client.swift
@@ -6,6 +6,7 @@ public typealias PreEventCallback = () async throws -> Void
 public enum ClientError: Error, CustomStringConvertible, LocalizedError {
 	case creationError(String)
 	case missingInboxId
+	case invalidInboxId(String)
 
 	public var description: String {
 		switch self {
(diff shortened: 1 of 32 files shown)


Sources/XMTPiOS/Client.swift
```diff
@@ -406,12 +410,11 @@ public final class Client {
...
 			allowReassignInboxId
-			? nil : try await inboxIdFromAddress(address: newAccount.address)
+			? nil : try await inboxIdFromIdentity(identity: newAccount.identity)
 
 		if allowReassignInboxId || (inboxId?.isEmpty ?? true) {
-			let signatureRequest = try await ffiAddWallet(
-				addressToAdd: newAccount.address.lowercased())
-
+			let signatureRequest = try await ffiAddIdentity(
+				identityToAdd: newAccount.identity)
```
cameronvoell, line 417: `testAddAccountsWithExistingInboxIds` in ClientTests was failing, looks like this fixes it:

```suggestion
			let signatureRequest = try await ffiAddIdentity(
				identityToAdd: newAccount.identity,
                                 allowReassignInboxId: allowReassignInboxId
                          )
```

cameronvoell: fixed in 7c38d2cafbdf27c271d229a2570aaaef4fcd9c0f

cameronvoell (approved):

Issue #474, opened 2025-03-05
Update Example App


nplasterer:
With the transition to V3 it would be great to update our example app to clearly display how to best integrator with our SDKs

**Core key concepts**
- V3 is all installation based that means conversations and messages do not persist between conversations unless message history is enabled and used correctly
- V3 is all local database driven that means new conversations and messages will not display unless sync is called or streaming is correctly performed

### MVP

- [ ] Show good proper client management for a random wallet (storing encryption keys and building clients)
- [x] Show correct usage of syncing
- [ ] Show correct usage of streaming
- [x] Show simple listing of conversations and messages
- [ ] Show simple creating conversations
- [x] Identity management showing useful and performant ways to display ENS names from inboxIds

### Future nice to haves

- [ ] Working push notifications example
- [ ] Message content type management
- [ ] Logging in with different wallets SCW, EOA, Passkey
- [ ] Backups
- [ ] Group management

PR #478, opened 2025-03-08
feat: scaffolding/nav for new example app

dmccartney:
This is the the first bits of #474


example/example/Database.swift
```diff
@@ -0,0 +1,216 @@
...
+
+/// Define the database schema and perform operations against it.
+actor Db {
+    
+    // This schema is defined in the series of @Model classes below.
+    static let schema = Schema([
+        Db.Conversation.self,
+        Db.Message.self,
+        Db.User.self
+    ])
```
dmccartney, line 13: 👀

Package.swift
```diff
@@ -17,6 +17,7 @@ let package = Package(
 		),
 	],
 	dependencies: [
+		// TODO: .package(name: "secp256k1", url: "https://github.com/GigaBitcoin/secp256k1.swift", .upToNextMinor(from: "0.10.0")),
```
nplasterer, line 20: We can't bring these back because of cocoapod issues. Can you remove these todos?

Package.swift
```diff
@@ -17,6 +17,7 @@ let package = Package(
...
 		.package(url: "https://github.com/tesseract-one/CSecp256k1.swift.git", from: "0.2.0"),
```
dmccartney, line 21: I'm leaving this dependency change commented out until a later PR where I can work through pod lint issues.

XMTP.podspec
```diff
@@ -22,6 +22,7 @@ Pod::Spec.new do |spec|
   spec.frameworks 		= "CryptoKit", "UIKit"
 
   spec.dependency 'CSecp256k1', '~> 0.2'
+  #TODO: spec.dependency 'secp256k1.swift', '~> 0.1'
```
nplasterer, line 25: Can't bring this back. Would also cause issues for out integrators if we changed our podspec would prefer if we just removed this todo

nplasterer (approved):

PR #479, opened 2025-03-10
Update Signer to be more universal

nplasterer:
Lets make the signer more universal so it can handle more signature types long term.


cameronvoell (approved):
🚀

PR #550, opened 2025-08-14
Fix retain cycle in Client

jarodl:
## Introduction 📟
`Client` kept a strong reference to `Conversations`, `PrivatePreferences`, and `XMTPDebugInformation`, all which kept a strong reference to `Client`. I also noticed that `XmtpApiClient` objects hung around because of `ApiClientCache` so I changed `ApiClientCache` so it holds weak references only.

## Purpose ℹ️ 
Fix retain cycle in `Client`

## Testing 🧪
Ran tests and checked the memory graph in the `example` app

### DIFF
--- a/Sources/XMTPiOS/Client.swift
+++ b/Sources/XMTPiOS/Client.swift
@@ -88,23 +88,23 @@ public struct ClientOptions {
 }
 
 actor ApiClientCache {
-	private var apiClientCache: [String: XmtpApiClient] = [:]
-	private var syncApiClientCache: [String: XmtpApiClient] = [:]
+    private var apiClientCache: NSMapTable<NSString, XmtpApiClient> = .weakToWeakObjects()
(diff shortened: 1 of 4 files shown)


cameronvoell:
@yewreeka I just did a dev release of these changes off latest `main` and stable libxmtp `main` using the branch [cv/jarod/client-retain-cycle-dev-release](https://github.com/xmtp/xmtp-ios/tree/cv/jarod/client-retain-cycle-dev-release)

Release version will be:
- [4.9.0-dev.fce3ad0](https://github.com/xmtp/xmtp-ios/releases/tag/4.9.0-dev.fce3ad0)
- [4.9.0-dev.fce3ad0-dylib](https://github.com/xmtp/xmtp-ios/releases/tag/4.9.0-dev.fce3ad0-dylib)

PR #553, opened 2025-08-14
Implement messagesV2

neekolas:
## Introduction 📟
- Introduces new `messagesV2` function which returns `DecodedMessageV2` results

## Purpose ℹ️ 
- Embeds replies and reactions into each result to avoid needing to keep separate lookup tables.

### DIFF
--- a/Sources/XMTPiOS/Libxmtp/DecodedMessageV2.swift
+++ b/Sources/XMTPiOS/Libxmtp/DecodedMessageV2.swift
@@ -7,9 +7,10 @@ public struct DecodedMessageV2: Identifiable {
     public var id: String {
         ffiMessage.id().toHex
     }
-
-    // Note: conversationId is not directly available in FfiDecodedMessage
-    // It would need to be passed in or derived from context
+    
+    public var conversationId: String {
+        ffiMessage.conversationId().toHex
+    }
 
     public var senderInboxId: InboxId {
         ffiMessage.senderInboxId()
@@ -34,9 +35,9 @@ public struct DecodedMessageV2: Identifiable {
         }
     }
 
-    // Topic generation requires conversationId which is not available in FfiDecodedMessage
+
     public var topic: String {
-        "" // Would need to be passed in or derived from context
+        Topic.groupMessage(conversationId).description
     }
 
     public var reactions: [DecodedMessageV2]? {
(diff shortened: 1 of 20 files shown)


Sources/XMTPiOS/Libxmtp/DecodedMessageV2.swift
```diff
@@ -0,0 +1,506 @@
...
+        "" // Would need to be passed in or derived from context
+    }
+
+    public var reactions: [DecodedMessageV2]? {
+        let reactionMessages = ffiMessage.reactions()
+        guard !reactionMessages.isEmpty else { return nil }
+        return reactionMessages.compactMap { DecodedMessageV2(ffiMessage: $0) }
+    }
+
+    public func content<T>() throws -> T {
```
neekolas, line 48: One thing that's controversial here is that we don't have an `encodedContent` field. I think that's a cleaner separation, and is certainly more memory efficient if we can throw out the original payload. Not sure who was using it. We pull out the content type, fallback, etc and have that available separately.

Sources/XMTPiOS/Libxmtp/DecodedMessageV2.swift
```diff
@@ -0,0 +1,506 @@
...
+            default:
+                // Unknown compression value, leave it unset
+                break
+            }
+        }
+        return encoded
+    }
+}
+
+// I included this for completeness in Rust, but I think it's unused except for really old groups.
```
neekolas, line 456: Should I just nuke this from Rust? I added it for completeness, but isn't it kinda dead and replaced with `GroupUpdated`? Would only be there for legacy reasons.

neekolas:
* **#560** <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-ios/560?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#553** <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-ios/553?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-ios/553?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

PR #564, opened 2025-09-05
Build xmtp ios without libxmtp-swift (including spm + cocoapod compatibility)

cameronvoell:


cameronvoell:
Verified that cocoapod published from here with no libxmtp-swift works fine with xmtp-react-native on a branch here:

https://github.com/xmtp/xmtp-react-native/compare/cv/verify-cocoa-pod-with-no-libxmtp-swift?expand=1

PR #621, opened 2025-11-19
Update to libxmtp 4.7.0-dev.f5bec47

neekolas:
### Upgrade to libxmtp 4.7.0-dev.f5bec47 and add sort and insertion-time filters to `Conversation.listMessages`, `Dm.listMessages`, and `Group.listMessages`
Add `sortBy`, `insertedAfterNs`, and `insertedBeforeNs` options to message listing APIs and expose `insertedAt`/`insertedAtNs` on decoded messages, backed by updated FFI bindings; bump package and podspec to the new binary release.

#### 📍Where to Start
Start with the new FFI options and enums in [xmtpv3.swift](https://github.com/xmtp/xmtp-ios/pull/621/files#diff-ec5124fcbfc3feda814ac7f92ee26fadffc570075abb8cc98d4a5de65a3c6df0), then review the `MessageSortBy` mapping in [DecodedMessage.swift](https://github.com/xmtp/xmtp-ios/pull/621/files#diff-96872bc3f891752324e391a9c2e014501ba307c648ebeb8b92509aaabb6bc049), and finally the updated list APIs in [Conversation.swift](https://github.com/xmtp/xmtp-ios/pull/621/files#diff-762ec7c3d6db84803270cf7dbc8426b70c8a716764921342fb61903384ee97d1), [Dm.swift](https://github.com/xmtp/xmtp-ios/pull/621/files#diff-c0f8e31a1bb57d01d49812b52b2db9459670c99095e042330e91a46b67c205c6), and [Group.swift](https://github.com/xmtp/xmtp-ios/pull/621/files#diff-1fffa18be2bf50b61ad120aae376479b48b94575bd42b97c772b8688acfe0f83).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized acd15ff. 8 files reviewed, 32 issues evaluated, 30 issues filtered, 1 comment posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>Sources/XMTPiOS/Dm.swift — 0 comments posted, 6 evaluated, 6 filtered</summary>

- [line 338](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Dm.swift#L338): `limit` is accepted without validation and converted directly via `options.limit = Int64(limit)`. Negative, zero, or excessively large values are not guarded. If the FFI expects a positive bound, this can lead to undefined or incorrect runtime behavior. Add validation (e.g., clamp to a reasonable maximum and enforce `limit > 0`) and return an error for invalid values. <b>[ Out of scope ]</b>
- [line 373](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Dm.swift#L373): Messages that fail to decode are silently dropped due to `compactMap`, yielding partial results without any visibility or error. This is a data-conversion boundary and should either surface an error, return a separate list of failures, or explicitly log/flag the discarded items. As written, callers cannot distinguish between “no messages” and “messages existed but failed to decode.” A safer approach is to use `map` with explicit handling, or collect errors alongside successes. <b>[ Out of scope ]</b>
- [line 413](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Dm.swift#L413): `limit` is accepted as `Int?` and converted to `Int64` without validation in both `Dm.messagesWithReactions` and `Group.messagesWithReactions`. A negative `limit` or an extremely large value can be forwarded to FFI, potentially causing backend errors or unexpected behavior. Validate that `limit` is positive and within a sensible upper bound before assigning `options.limit`. <b>[ Low confidence ]</b>
- [line 417](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Dm.swift#L417): In `Dm.messagesWithReactions`, `options.deliveryStatus` is set via `deliveryStatus.toFfi()` while `Group.messagesWithReactions` maps `.all` to `nil` explicitly. This inconsistency can yield different filtering behavior between DMs and Groups when `deliveryStatus` is `.all` if `toFfi()` does not return `nil` for `.all`. This breaks contract parity across conversation types and can lead to unexpected result sets. <b>[ Low confidence ]</b>
- [line 419](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Dm.swift#L419): Both `Dm.messagesWithReactions` and `Group.messagesWithReactions` coerce a `nil` `direction` into `FfiDirection.descending`. If `nil` is intended to mean ‘unspecified’ (i.e., let backend default or preserve insertion order), this forces a sort and may change results ordering compared to passing `nil`. Given the parameter type is `SortDirection?`, callers can pass `nil`, but it never reaches FFI as `nil`. <b>[ Low confidence ]</b>
- [line 537](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Dm.swift#L537): Both `Dm.enrichedMessages` and `Group.enrichedMessages` return results using `compactMap` when constructing `DecodedMessageV2`, i.e., `return try await ffiConversation.findEnrichedMessages(opts: options).compactMap { ffiDecodedMessage in DecodedMessageV2(ffiMessage: ffiDecodedMessage) }` and similarly for `ffiGroup`. Using `compactMap` silently drops any messages whose decoding fails (where `DecodedMessageV2(...)` returns `nil`), causing silent data loss and potentially breaking caller expectations for message counts or completeness. If decoding can fail, this should surface an explicit error or include placeholders with status, rather than silently omitting entries. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Sources/XMTPiOS/Group.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 571](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Group.swift#L571): Using `compactMap` when mapping `ffiGroup.findMessages(opts:)` results to `DecodedMessage` silently drops any messages for which `DecodedMessage.create(ffiMessage:)` returns `nil`. This causes undetected data loss at the data-conversion boundary: callers get fewer messages without any error or indication that decoding failed for some items. If decoding can fail for legitimate inputs, you should either surface errors (e.g., return a result that includes failures), log/report them, or use a fall back to preserve contract visibility rather than silently filtering. <b>[ Low confidence ]</b>
- [line 724](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Group.swift#L724): In `Group.enrichedMessages`, the same silent data loss occurs due to `compactMap` when constructing `DecodedMessageV2`: `return try await ffiGroup.findEnrichedMessages(opts: options).compactMap { ffiDecodedMessage in DecodedMessageV2(ffiMessage: ffiDecodedMessage) }`. Any undecodable message is dropped without error or indication, potentially violating contract parity and caller expectations. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Sources/XMTPiOS/Libxmtp/DecodedMessage.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 202](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/DecodedMessage.swift#L202): In `public static func create(ffiMessage: FfiMessageWithReactions) -> DecodedMessage?`, a single failure while decoding any reaction causes the entire parent message to be dropped. Specifically, the `map` over `ffiMessage.reactions` at lines 216–226 throws if any `reaction` content fails to decode, which is then caught by the surrounding `do/catch` at lines 202–235, resulting in `return nil` for the whole message. This silently discards an otherwise valid `ffiMessage.message` and all successfully decoded reactions. A safer behavior would be to skip or flag only the malformed reactions and still return the parent `DecodedMessage` with the decodable subset of `childMessages`. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Sources/XMTPiOS/Libxmtp/DecodedMessageV2.swift — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 143](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/DecodedMessageV2.swift#L143): `mapContent(.custom)` calls `Client.codecRegistry.find(for: encoded.type)` even when `encoded.type` may be unset because `mapFfiEncodedContent(_:)` only sets `encoded.type` if `ffiContent.typeId` is non-nil. If `typeId` is nil, `encoded.type` may be a default/empty value or nil (depending on `EncodedContent`'s definition), which can cause a lookup failure or unexpected behavior (e.g., crash if `find` expects a valid type, or decoding with the wrong codec). Guard for missing `typeId` before attempting codec lookup, or provide a safe default/explicit error. <b>[ Low confidence ]</b>
- [line 230](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/DecodedMessageV2.swift#L230): `determineContentType(from:)` returns `ContentTypeText` for `.custom` when `typeId` is missing (lines 230–232). This mislabels content and creates contract mismatch: callers observing `contentType` will see `text` even though the body is `.custom`. It can lead to incorrect handling downstream and violates contract parity with `mapMessageBody(.custom)` which attempts codec-based decoding. <b>[ Low confidence ]</b>
- [line 239](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/DecodedMessageV2.swift#L239): `mapReaction(_:)` maps `action` as `reactionPayload.action == .added ? .added : .removed` (line 239). Any non-`.added` value, including potential `.unknown` or future enum cases, is coerced to `.removed`, silently misclassifying reactions. Use an explicit `switch` to preserve semantics and handle unknown values distinctly. <b>[ Low confidence ]</b>
- [line 292](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/DecodedMessageV2.swift#L292): `mapMultiRemoteAttachment(_:)` forwards `info.scheme` directly (line 292) while `mapRemoteAttachment(_:)` normalizes schemes via `mapRemoteAttachmentScheme(_:)` to ensure a valid enum with default `.https`. This inconsistency can yield invalid scheme values for multi-attachments (e.g., unknown strings), causing downstream errors or contract mismatch. Normalize `info.scheme` the same way. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Sources/XMTPiOS/Libxmtp/xmtpv3.swift — 0 comments posted, 9 evaluated, 8 filtered</summary>

- [line 2552](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/xmtpv3.swift#L2552): `try! await uniffiRustCallAsync(...)` will crash the process if `uniffiRustCallAsync` throws (e.g., if the Rust future returns an error or `liftFunc` fails). This is a normal runtime path, not an initialization invariant, and no error handler is provided (`errorHandler: nil`). Replace `try!` with `try` and propagate/handle the error so the app does not terminate unexpectedly. <b>[ Low confidence ]</b>
- [line 2558](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/xmtpv3.swift#L2558): Potential retain cycle/leak: the `callback` lowered via `FfiConverterTypeFfiMessageDeletionCallback_lower(callback)` may be retained by the underlying stream while the owner retains the returned `FfiStreamCloser`. If the `callback` strongly captures the owner, a reference cycle can form (`owner -> stream closer -> FFI-held callback -> owner`), preventing deallocation until the stream is explicitly closed. Document the capture guidelines or use weak/unowned captures and ensure the `FfiStreamCloser` is always closed to break the cycle. <b>[ Low confidence ]</b>
- [line 3653](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/xmtpv3.swift#L3653): Potential resource/state leak: `lower(_:)` calls `handleMap.insert(obj:)` but this converter provides no corresponding removal/cleanup on any exit path (success or failure). Repeated calls will grow `handleMap` and retain `FfiMessageDeletionCallback` instances, with no visible lifecycle pairing. <b>[ Low confidence ]</b>
- [line 3654](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/xmtpv3.swift#L3654): `lower(_:)` may crash via `fatalError("Cast to UnsafeMutableRawPointer failed")` if `UniffiHandleMap.insert(obj:)` returns a handle value of `0`, because `UnsafeMutableRawPointer(bitPattern: 0)` is `nil`. This occurs after the object is inserted into `handleMap`, leaving a mutated state with no paired cleanup on this path. <b>[ Low confidence ]</b>
- [line 3667](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/xmtpv3.swift#L3667): `read(from:)` accepts any `UInt64` and converts it to a pointer, throwing only on `0`/nil. It then `lift`s the pointer without validating it corresponds to a previously inserted handle. If an invalid or stale handle value is provided (e.g., from Rust or corrupted data), subsequent use of the lifted callback may fail or crash due to missing map entries or misuse, with no graceful fallback here. <b>[ Low confidence ]</b>
- [line 6576](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/xmtpv3.swift#L6576): `FfiDecodedMessageMetadata` declares `Hashable` conformance while providing a custom `==` that compares a specific subset of properties (`id`, `sentAtNs`, `kind`, `senderInstallationId`, `senderInboxId`, `contentType`, `conversationId`, `insertedAtNs`). If `Hashable` is synthesized (no explicit `hash(into:)` present), the synthesized hash will include all stored properties. This can violate the Hashable contract if there are stored properties not covered by `==`: two instances may be considered equal by `==` yet produce different hash values, causing incorrect behavior in hashed collections (e.g., `Set`, `Dictionary`). To fix, either ensure `==` and `hash(into:)` operate over the exact same set of properties, or allow the compiler to synthesize both Equatable and Hashable using identical properties. <b>[ Low confidence ]</b>
- [line 6622](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/xmtpv3.swift#L6622): Parsing advances `buf.offset` incrementally via multiple `read(from: &buf)` calls. If any intermediate read fails (e.g., due to insufficient data or malformed field), the function throws but leaves `buf.offset` partially advanced, corrupting the caller’s parsing state. To preserve invariants, parsing must be transactional: either use a local copy of `offset` and commit only on success, or ensure each `read` returns decoded data without mutating `buf` until the whole decode succeeds. <b>[ Low confidence ]</b>
- [line 6632](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOS/Libxmtp/xmtpv3.swift#L6632): The newly added field `insertedAtNs: FfiConverterInt64.read(from: &buf)` changes the decode contract by requiring an additional trailing `Int64`. If older serialized payloads omit this field, decoding will attempt to read past the available data and throw, breaking backward compatibility. There is no versioning, presence flag, or default-value handling to safely decode older messages. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Sources/XMTPiOSDynamic/Libxmtp/xmtpv3.swift — 1 comment posted, 10 evaluated, 9 filtered</summary>

- [line 101](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOSDynamic/Libxmtp/xmtpv3.swift#L101): Potential integer overflow when computing `reader.offset + count`. If the sum overflows `Int`, the program traps at runtime during the addition, prior to any guard or error handling. <b>[ Low confidence ]</b>
- [line 101](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOSDynamic/Libxmtp/xmtpv3.swift#L101): `count` is not validated to be non-negative before constructing `range = reader.offset ..< (reader.offset + count)`. If `count < 0`, creating this range will trap at runtime due to an invalid range (upperBound < lowerBound) before the guard executes. <b>[ Low confidence ]</b>
- [line 102](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOSDynamic/Libxmtp/xmtpv3.swift#L102): `reader.offset` is not validated to be within `0...reader.data.count`. If `offset` is negative, the guard `reader.data.count >= range.upperBound` can still pass, and `Data.copyBytes(to:from:)` will be invoked with a range whose lowerBound is < 0, causing a runtime precondition failure. <b>[ Low confidence ]</b>
- [line 302](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOSDynamic/Libxmtp/xmtpv3.swift#L302): In `CALL_ERROR` when an `errorHandler` is provided, the code throws the handler's error without deallocating `callStatus.errorBuf`. If the handler does not explicitly take ownership and deallocate the buffer, this causes a memory leak. To be safe, either always deallocate after the handler converts the buffer (e.g., by ensuring the handler consumes and deallocates), or deallocate here after obtaining the error value. <b>[ Low confidence ]</b>
- [line 313](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOSDynamic/Libxmtp/xmtpv3.swift#L313): In `CALL_UNEXPECTED_ERROR` when `callStatus.errorBuf.len > 0`, the code throws `UniffiInternalError.rustPanic(FfiConverterString.lift(callStatus.errorBuf))` without explicitly deallocating `errorBuf`. If `FfiConverterString.lift` does not deallocate the buffer or if it throws before deallocation, the buffer leaks. Ensure `errorBuf` is always freed exactly once on all exit paths, including when `lift` throws (e.g., by first converting safely, then deallocating in a `defer`, and falling back to a default message if conversion fails). <b>[ Low confidence ]</b>
- [line 314](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOSDynamic/Libxmtp/xmtpv3.swift#L314): In `CALL_UNEXPECTED_ERROR` when converting the panic message, `FfiConverterString.lift(callStatus.errorBuf)` may throw (e.g., invalid encoding). The current code propagates that conversion error instead of reliably surfacing a Rust panic. This changes the externally visible error contract and may hide the original panic. Provide a fallback: if conversion fails, deallocate the buffer and throw `UniffiInternalError.rustPanic("Rust panic")` or a safe-decoded message. <b>[ Low confidence ]</b>
- [line 323](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOSDynamic/Libxmtp/xmtpv3.swift#L323): In the `default` case, the code throws `UniffiInternalError.unexpectedRustCallStatusCode` without deallocating `callStatus.errorBuf`. If the buffer was populated by the callee, this results in a memory leak. Always free the buffer before throwing on unknown status codes. <b>[ Low confidence ]</b>
- [line 334](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOSDynamic/Libxmtp/xmtpv3.swift#L334): On the success path, `callStatus.pointee.code` is never set to a success value (e.g., `CALL_SUCCESS`). If the caller reuses a `RustCallStatus` instance from a prior error, the stale `code` can remain non-success, causing the FFI caller to misinterpret this call as an error even though it succeeded. <b>[ Low confidence ]</b>
- [line 334](https://github.com/xmtp/xmtp-ios/blob/acd15ff958587db284f0b66de11e060d1e6d50bb/Sources/XMTPiOSDynamic/Libxmtp/xmtpv3.swift#L334): On the success path, `callStatus.pointee.errorBuf` is not cleared/reset. If a previous error populated `errorBuf`, the stale pointer/message can persist on success, leading to incorrect error reporting and potential memory/resource leaks if the caller frees the stale buffer based on the non-success state. <b>[ Low confidence ]</b>
</details>

</details>

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -25,8 +25,8 @@ let package = Package(
 	targets: [
 		.binaryTarget(
 			name: "LibXMTPSwiftFFI",
-			url: "https://github.com/xmtp/libxmtp/releases/download/swift-bindings-1.6.1-rc2.5597847/LibXMTPSwiftFFI.zip",
-			checksum: "760d953d565f7070e76941b955647118f21950f3877d80d1ae9395f88bef0d8c"
+			url: "https://github.com/xmtp/libxmtp/releases/download/swift-bindings-1.7.0-dev.f5bec47/LibXMTPSwiftFFI.zip",
+			checksum: "29d7aef74978d8c2c45a028ce80e7bce92b4a5b9b2470606b7798632bbf9bd6b"
 		),
 		.target(
 			name: "XMTPiOS",
--- a/Sources/XMTPiOS/Conversation.swift
+++ b/Sources/XMTPiOS/Conversation.swift
@@ -280,7 +280,10 @@ public enum Conversation: Identifiable, Equatable, Hashable {
 		direction: SortDirection? = .descending,
 		deliveryStatus: MessageDeliveryStatus = .all,
 		excludeContentTypes: [StandardContentType]? = nil,
-		excludeSenderInboxIds: [String]? = nil
+		excludeSenderInboxIds: [String]? = nil,
+		sortBy: MessageSortBy? = nil,
+		insertedAfterNs: Int64? = nil,
+		insertedBeforeNs: Int64? = nil
 	) async throws -> [DecodedMessage] {
 		switch self {
 		case let .group(group):
@@ -288,14 +291,20 @@ public enum Conversation: Identifiable, Equatable, Hashable {
 				beforeNs: beforeNs, afterNs: afterNs, limit: limit,
 				direction: direction, deliveryStatus: deliveryStatus,
 				excludeContentTypes: excludeContentTypes,
-				excludeSenderInboxIds: excludeSenderInboxIds
+				excludeSenderInboxIds: excludeSenderInboxIds,
+				sortBy: sortBy,
+				insertedAfterNs: insertedAfterNs,
+				insertedBeforeNs: insertedBeforeNs
 			)
 		case let .dm(dm):
 			return try await dm.messages(
 				beforeNs: beforeNs, afterNs: afterNs, limit: limit,
 				direction: direction, deliveryStatus: deliveryStatus,
 				excludeContentTypes: excludeContentTypes,
-				excludeSenderInboxIds: excludeSenderInboxIds
+				excludeSenderInboxIds: excludeSenderInboxIds,
+				sortBy: sortBy,
+				insertedAfterNs: insertedAfterNs,
+				insertedBeforeNs: insertedBeforeNs
 			)
 		}
 	}
@@ -326,7 +335,10 @@ public enum Conversation: Identifiable, Equatable, Hashable {
 		direction: SortDirection? = .descending,
 		deliveryStatus: MessageDeliveryStatus = .all,
 		excludeContentTypes: [StandardContentType]? = nil,
-		excludeSenderInboxIds: [String]? = nil
+		excludeSenderInboxIds: [String]? = nil,
+		sortBy: MessageSortBy? = nil,
+		insertedAfterNs: Int64? = nil,
+		insertedBeforeNs: Int64? = nil
 	) async throws -> [DecodedMessage] {
 		switch self {
 		case let .group(group):
@@ -334,14 +346,20 @@ public enum Conversation: Identifiable, Equatable, Hashable {
 				beforeNs: beforeNs, afterNs: afterNs, limit: limit,
 				direction: direction, deliveryStatus: deliveryStatus,
 				excludeContentTypes: excludeContentTypes,
-				excludeSenderInboxIds: excludeSenderInboxIds
+				excludeSenderInboxIds: excludeSenderInboxIds,
+				sortBy: sortBy,
+				insertedAfterNs: insertedAfterNs,
+				insertedBeforeNs: insertedBeforeNs
 			)
 		case let .dm(dm):
 			return try await dm.messagesWithReactions(
 				beforeNs: beforeNs, afterNs: afterNs, limit: limit,
 				direction: direction, deliveryStatus: deliveryStatus,
 				excludeContentTypes: excludeContentTypes,
-				excludeSenderInboxIds: excludeSenderInboxIds
+				excludeSenderInboxIds: excludeSenderInboxIds,
+				sortBy: sortBy,
+				insertedAfterNs: insertedAfterNs,
+				insertedBeforeNs: insertedBeforeNs
 			)
 		}
 	}
@@ -353,7 +371,10 @@ public enum Conversation: Identifiable, Equatable, Hashable {
 		direction: SortDirection? = .descending,
 		deliveryStatus: MessageDeliveryStatus = .all,
 		excludeContentTypes: [StandardContentType]? = nil,
-		excludeSenderInboxIds: [String]? = nil
+		excludeSenderInboxIds: [String]? = nil,
+		sortBy: MessageSortBy? = nil,
+		insertedAfterNs: Int64? = nil,
+		insertedBeforeNs: Int64? = nil
 	) async throws -> [DecodedMessageV2] {
 		switch self {
 		case let .group(group):
@@ -361,14 +382,20 @@ public enum Conversation: Identifiable, Equatable, Hashable {
 				beforeNs: beforeNs, afterNs: afterNs, limit: limit,
 				direction: direction, deliveryStatus: deliveryStatus,
 				excludeContentTypes: excludeContentTypes,
-				excludeSenderInboxIds: excludeSenderInboxIds
+				excludeSenderInboxIds: excludeSenderInboxIds,
+				sortBy: sortBy,
+				insertedAfterNs: insertedAfterNs,
+				insertedBeforeNs: insertedBeforeNs
 			)
 		case let .dm(dm):
 			return try await dm.enrichedMessages(
 				beforeNs: beforeNs, afterNs: afterNs, limit: limit,
 				direction: direction, deliveryStatus: deliveryStatus,
 				excludeContentTypes: excludeContentTypes,
-				excludeSenderInboxIds: excludeSenderInboxIds
+				excludeSenderInboxIds: excludeSenderInboxIds,
+				sortBy: sortBy,
+				insertedAfterNs: insertedAfterNs,
+				insertedBeforeNs: insertedBeforeNs
 			)
 		}
 	}
@@ -378,20 +405,26 @@ public enum Conversation: Identifiable, Equatable, Hashable {
 		afterNs: Int64? = nil,
 		deliveryStatus: MessageDeliveryStatus = .all,
 		excludeContentTypes: [StandardContentType]? = nil,
-		excludeSenderInboxIds: [String]? = nil
+		excludeSenderInboxIds: [String]? = nil,
+		insertedAfterNs: Int64? = nil,
+		insertedBeforeNs: Int64? = nil
 	) throws -> Int64 {
 		switch self {
 		case let .group(group):
 			return try group.countMessages(
 				beforeNs: beforeNs, afterNs: afterNs, deliveryStatus: deliveryStatus,
 				excludeContentTypes: excludeContentTypes,
-				excludeSenderInboxIds: excludeSenderInboxIds
+				excludeSenderInboxIds: excludeSenderInboxIds,
+				insertedAfterNs: insertedAfterNs,
+				insertedBeforeNs: insertedBeforeNs
 			)
 		case let .dm(dm):
 			return try dm.countMessages(
 				beforeNs: beforeNs, afterNs: afterNs, deliveryStatus: deliveryStatus,
 				excludeContentTypes: excludeContentTypes,
-				excludeSenderInboxIds: excludeSenderInboxIds
+				excludeSenderInboxIds: excludeSenderInboxIds,
+				insertedAfterNs: insertedAfterNs,
+				insertedBeforeNs: insertedBeforeNs
 			)
 		}
 	}
--- a/Sources/XMTPiOS/Dm.swift
+++ b/Sources/XMTPiOS/Dm.swift
@@ -308,7 +308,10 @@ public struct Dm: Identifiable, Equatable, Hashable {
 		direction: SortDirection? = .descending,
 		deliveryStatus: MessageDeliveryStatus = .all,
 		excludeContentTypes: [StandardContentType]? = nil,
-		excludeSenderInboxIds: [String]? = nil
+		excludeSenderInboxIds: [String]? = nil,
+		sortBy: MessageSortBy? = nil,
+		insertedAfterNs: Int64? = nil,
+		insertedBeforeNs: Int64? = nil
 	) async throws -> [DecodedMessage] {
 		var options = FfiListMessagesOptions(
 			sentBeforeNs: nil,
@@ -318,7 +321,10 @@ public struct Dm: Identifiable, Equatable, Hashable {
 			direction: nil,
 			contentTypes: nil,
 			excludeContentTypes: nil,
-			excludeSenderInboxIds: nil
+			excludeSenderInboxIds: nil,
+			sortBy: nil,
+			insertedAfterNs: nil,
+			insertedBeforeNs: nil
 		)
 
 		if let beforeNs {
@@ -360,6 +366,9 @@ public struct Dm: Identifiable, Equatable, Hashable {
 		options.direction = direction
 		options.excludeContentTypes = excludeContentTypes
 		options.excludeSenderInboxIds = excludeSenderInboxIds
+		options.sortBy = sortBy?.toFfi()
+		options.insertedAfterNs = insertedAfterNs
+		options.insertedBeforeNs = insertedBeforeNs
 
 		return try await ffiConversation.findMessages(opts: options).compactMap {
 			ffiMessage in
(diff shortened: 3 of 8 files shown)


neekolas:
* **#621** <a href="https://app.graphite.com/github/pr/xmtp/xmtp-ios/621?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtp/xmtp-ios/621?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

PR #669, opened 2026-02-03
Get the reference docs generating

jhaaaa:
## Introduction 📟
The SDK reference docs for this repo stopped generating in Dec 2025. This PR enables them to generate again.

## Purpose ℹ️ 
Enable SDK reference docs to start generating again.

## Scope 🔭
- Updates .github/workflows/docs.yml to enable reference docs to start generating again.
- Specifically, the PR switches from using Jazzy to DocC for ref doc generation. Per Claude Code: Jazzy relies on SourceKit for module discovery, which fails on Swift packages with binary framework dependencies. Since this project uses LibXMTPSwiftFFI.xcframework (the Rust FFI bindings), Jazzy cannot resolve the XMTPiOS module and documentation generation fails silently. DocC is Apple's official documentation tool, is already configured as a dependency in Package.swift (swift-docc-plugin), and works seamlessly with binary targets. This change also simplifies the CI workflow by removing the Ruby/Bundler setup previously required for Jazzy.

Is there a requirement that we use Jazzy?


jhaaaa:
Moving to libxmtp.

Issue #373, opened 2024-07-24
Feature request: Remove `@testable` from tests unless it is explicitly needed


cameronvoell:
### Is your feature request related to a problem?

There are some spots in our tests where we are utilizing internal only properties that incorrectly demonstrate how to use the SDK. Removing `@testable` import of XMTPiOS from our tests will help us catch this in tests.

PR #670, opened 2026-02-20
Remove unneeded `@testable` imports

shyamkumar703:
## Introduction 📟

Replaces `@testable import XMTPiOS` with `import XMTPiOS` where possible; so tests are testing the public SDK API instead of internal fields/functions

## Purpose ℹ️ 

Test the public API instead of internal fields/functions

## Scope 🔭

Test cases

### DIFF
--- a/Sources/XMTPTestHelpers/TestHelpers.swift
+++ b/Sources/XMTPTestHelpers/TestHelpers.swift
@@ -108,4 +108,15 @@
 			try await Fixtures(clientOptions: clientOptions)
 		}
 	}
+
+// INFO: - copied from Date.swift
+public extension Date {
+	var millisecondsSinceEpoch: Double {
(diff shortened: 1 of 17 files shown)


shyamkumar703:
Proposed fix for https://github.com/xmtp/xmtp-ios/issues/373

Sources/XMTPTestHelpers/TestHelpers.swift
```diff
@@ -104,4 +108,15 @@
 			try await Fixtures(clientOptions: clientOptions)
 		}
 	}
+
+// INFO: - copied from Date.swift
+public extension Date {
```
shyamkumar703, line 113: Copied wholesale from `Date.swift` - a `Common` module with shared utilities across the test/source targets might be a better pattern here but seems overkill for _just_ this extension.

